PR #8, opened 2017-10-25
Added Contributors to Package.json

hmtinc:
- Added contributors

### DIFF
--- a/package.json
+++ b/package.json
@@ -124,11 +124,28 @@
   },
   "browserslist": "last 3 versions, >1%",
   "contributors": [
-    "Dylan Fong <dylanfong.ut@gmail.com>",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -122,5 +122,13 @@
   "engines": {
     "node": ">=6.3.0"
   },
-  "browserslist": "last 3 versions, >1%"
+  "browserslist": "last 3 versions, >1%",
+  "contributors": [
```
d2fong: It is probably better if we do it like cytoscape.js:

https://github.com/cytoscape/cytoscape.js/blob/master/package.json#L19

where names, emails, links have their own fields for each contributor

hmtinc: I updated the file so emails, names, and links have their own fields.

PR #23, opened 2017-10-25
Fixed lazyload.js to have 2 spaces instead of 4 for indentation

gaelder:


maxkfranz:
What about the other files?

gaelder:
I went through the files again. I found two that had one small section of indented with 4 spaces instead of 2. I've created a new pull request.

PR #40, opened 2017-10-26
Updated route names

jonahdlin:
Changed route names to dash case and lowercase


jonahdlin:
Fixed in #40

d2fong:
I also noticed there is a comment header in this file.  Can you remove it?

jonahdlin:
Yup, it's done.

PR #41, opened 2017-10-26
Renamed files in the database folder to not use DB or database. Fixed…

gaelder:
… all import statement as well


d2fong:
This is a good start, and I think 'query' is a better name than 'accessDB', but now we will have to re-arrange some of the functions in query.  

Generally, the name 'query' is associated with read only access, but there are some functions such as 'updateGraph', or 'saveLayout' that do more than query the db (i.e update/save).  

It would be nice to create a new file like 'update.js' that contained the updateGraph and saveLayout functions.  

I am not sure if 'updateVersion' is that descriptive.  Can you elaborate on what the purpose of this file is?  Is this code for a batch job, i.e run this code for each pathway?

d2fong:
@gaelder I am going to merge, but I created a related to organizing the db files.

Issue #21, opened 2017-10-25
Use ES6 implicit values where the object keys match the variable name


d2fong:
E.g. 

https://github.com/PathwayCommons/app-ui/blob/202e7685e41ae43f7150e36dd4b3f13b5c622f24/src/server/graph-generation/cytoscapeJson.js#L36-L42

Where you do `{ foo: foo }`, you can just type `{ foo }`.  It makes it easier to read.

jonahdlin:
Last of this issue resolved in #60

Issue #18, opened 2017-10-25
Promise chains should be flat -- not embedded like a callback pyramid


d2fong:
E.g. 

https://github.com/PathwayCommons/app-ui/blob/202e7685e41ae43f7150e36dd4b3f13b5c622f24/src/server/graph-generation/cytoscapeJson.js#L17-L48

You do

```js
(
Promise.resolve( getPromise1() )
.then( () => getPromise2() )
.then( () => getPromise3() )
);
```

Not

```js
getPromise1().then( () => {
  return getPromise2().then( () => {
    return getPromise3();
  } );
} );
```

Put the `Promise.resolve()` first so that if `getPromise1()` throws then you can `.catch()` it rather than `catch`ing it with a block.

PR #48, opened 2017-10-26
Reformatted Chained Promises and Module.Export Blocks

hmtinc:
- #18 Reformatted the chained promises to not be in a pyramid form, and alot more easier to read
- #21 Used ES6 implicit values to remove repeated references from objects and module.export blocks

Resolved following issues for graph-generation


src/server/graph-generation/treeTraversal.js
```diff
@@ -117,11 +117,11 @@ function searchForExactNodeWithObjectResult(subtree, key) {
...
-  searchMultiple: searchMultiple,
-  searchOne: searchOne,
-  searchTree: searchTree
+  searchForExactNode,
+  searchForExactNodeWithObjectResult,
+  searchForFirst,
+  searchForNode,
+  searchMultiple,
+  searchOne,
+  searchTree,
```
maxkfranz: Watch for trailing commas.  ESLint should pick these up in your editor.

src/server/graph-generation/cytoscapeJson.js
```diff
@@ -8,34 +8,28 @@ const metadataMapperPC2 = require('./metadataMapperPC2');
...
+      return pcServices.traversePC2(uri, 'Entity/organism/displayName');
+    })
+    .then(data => {
+      organism = data.traverseEntry[0].value;
+      return {
+        comments,
+        dataSource,
+        title,
+        organism
+      }
```
maxkfranz: Much better!  The next step to make things even faster and easier to read is to use short, arrow functions.

Something like:

```js
let getValue = data => data.where.the.field.is;
let get = name => getFromPc( name ).then( getValue );
```

Then you do a bunch in parallel:

```js
let foo, bar;

Promise.all([
  get('x/y').then( v => foo = v ),
  get('a/b').then( v => bar = v )
]);
```

If you have N things that you're getting, you can be up to N times faster this way, and it's more clear and concise.

hmtinc:
Huh, that does make it easier to read. I've applied the changes.

PR #52, opened 2017-10-26
JSX converted to hyperscript. Removed redundant files.

jonahdlin:


src/client/features/view/layout/coseBilkent.js
```diff
@@ -1,13 +0,0 @@
-const coseBilkentOpts = {
-  name: 'cose-bilkent',
-  fit: true,
```
d2fong: Did you still need these layout files?

jonahdlin: They were redundant. I must have quickly copied them at some point, but they exist elsewhere. I just changed the reference and deleted the copies.

d2fong: Oh okay.

src/client/features/admin/index.js
```diff
@@ -7,7 +7,7 @@ const classNames = require('classnames');
 
 const Icon = require('../../common/components').Icon;
 const PathwayCommonsService = require('../../services').PathwayCommonsService;
-const CDC = require('../../services/index.js').CDC;
+const CDC = require('../../services/index').CDC;
```
d2fong: @jonahdlin  @gaelder @hmtinc if you are requiring directories and there is an index.js file in that directory, it requires that automatically.

e.g
```js
const CDC = require('../../services/index').CDC;
// is equivalent to
const CDC = require('../../services/').CDC;
```

src/client/features/view/components/graph/index.js
```diff
@@ -62,23 +63,20 @@ class Graph extends React.Component {
...
-            ref={dom => this.graphDOM = dom}
-            style={{
-              width: this.state.width,
-              height: this.state.height
-            }}
-          />
-        </div>
+        h('div.Graph', [
+          h('div', {
+            'id': this.state.graphId,
```
d2fong: @hmtinc @jonahdlin @gaelder 

Pretty sure you don't need to have object keys as strings.  Maybe if the key shares a keyword like class, but in general you shouldn't have to write them as strings.

jonahdlin:
@d2fong Changed the strings in the object keys to plain keys. I also removed `index` from require statements and redid module.export statements to be cleaner.

PR #58, opened 2017-10-27
Redid batch DB populate job and removed async dependency

jonahdlin:
Replaced `async.whilst` with `bluebird.map`. I had to remove functionality to do this. The batch job no longer retries URIs that don't work with the JSONLD parse method. Since JSONLD is stable and if it fails, it's likely the other methods would fail as well.


jonahdlin:
This currently works quite poorly for a large amount of pathways. However, with plans for lazy loading, it's unlikely this batch job will be maintained after we leave anyway.

d2fong:
Will do a more in depth review today

d2fong:
Merge conflict in createTables.js:
```js
<<<<<<< batchJobs
r.connect( {host: config.ip, port: config.port}, function(err, conn) {
  if (err) throw err;
  connection = conn;
  createTables('metadataTestMk2', [
    'graph',
    'version',
    'layout',
    'layout_cache'
  ], true);
});
=======
function createTable(dbName, tableName, connection) {
  return r.db(dbName).tableCreate(tableName).run(connection);
}
>>>>>>> master
```

What is the correct resolution?

gaelder:
@d2fong keep the changes from master (function createTable)

PR #60, opened 2017-10-27
All services moved to src/service

jonahdlin:
References to these services are fixed wherever I could find them


jonahdlin:
Will redo and reopen this later. Pulling now would cause too many merge conflicts.

Issue #38, opened 2017-10-25
Socket / http code duplication


d2fong:
If you want to be able to do the same things with an HTTP request or a socket, then create functions that do the work that both the http route code and the socket code call.  There shouldn't be code duplication just because of the medium.

Issue #31, opened 2017-10-25
Modules in node generally should not execute things at the top-level


d2fong:
E.g.

https://github.com/PathwayCommons/app-ui/blob/202e7685e41ae43f7150e36dd4b3f13b5c622f24/src/server/database/createDatabase.js?utf8=%E2%9C%93#L13

If you want to create a command script, then put it in a separate top-level dir named `scripts`.

Or put all the functionality in functions and export them so anything can use the code.

E.g.

https://github.com/PathwayCommons/app-ui/blob/202e7685e41ae43f7150e36dd4b3f13b5c622f24/src/server/routes/databaseRoutes.js#L34

You should call `connect()` in each function that needs a connection and the `accessDB` object could do caching internally if you want.

PR #62, opened 2017-10-30
Route dedup

gaelder:
#38
One instance of #31


d2fong:
Will do a more in depth review today

Issue #32, opened 2017-10-25
DB init should not be a command script


d2fong:
Creating the DB should not be a batch operation or a script, in general.   Your top-level app code in node should call a function that you define that checks whether the db exists.  If not, it creates it and then resolves the promise.  That way it's automatic.

PR #63, opened 2017-10-30
Check for correct database instantiation on server startup

gaelder:
#32


d2fong:
Will do more in depth review today

src/server/database/config.js
```diff
@@ -1,5 +1,6 @@
 module.exports = {
-  databaseName : 'layouts',
+  databaseName : 'halfBaked',
```
d2fong: @hmtinc @gaelder the current config for the db is:
```js
  databaseName : 'metadataTestMk2',
  ip: '127.0.0.1',
```

Which one should be kept?

gaelder: Oops, my changes there should have been reverted. It should be 
  databaseName : 'metadataTestMk2',
  ip: '127.0.0.1',

PR #68, opened 2017-10-31
Pure CSS reorganize

jonahdlin:
CSS reorganization. This one should only have CSS changes.


jonahdlin:
pathways.txt should be deleted anyway, since it was mistakenly pushed in the first place

src/styles/features/view/menuBar.css
```diff
@@ -33,6 +36,7 @@
 .title-container {
   padding-left: calc(1em - var(--icon-size) / 3);
   padding-right: 1em;
+
   color: var(--midnight-blue);
```
maxkfranz: Var names should be semantic, not the name of the colour.  What if in future you want to change the variable value and it's not blue?

jonahdlin: Names updated. They no longer use the name of the colour, instead using a descriptor of where the colour should be used.

jonahdlin:
Redone in #108

Issue #134, opened 2017-11-03
Format comments in sidebar metadata viewer


hmtinc:
Currently the comments for each node, which are presented in the sidebar, are displayed as a comma separated list. This makes it hard to differ between different comments. 

Each comment should be separated by a line or blank space.

PR #150, opened 2017-11-08
Comment and Additional Information Formatting

hmtinc:
#134  - I added additional formatting to the data within the side bar. All the values that have more then one entry are in a list.


src/client/features/view/components/tooltips/metadataTip.js
```diff
@@ -22,15 +22,20 @@ class MetadataTip {
 
     //Get HTML for current data pair based on key value
     if (key === 'Standard Name') {
-      return h('div.fake-paragraph', h('div.field-name', 'Approved Name: '), pair[1].toString());
+      return this.makeTooltipItem(pair[1], 'Approved Name: ');
```
d2fong: Is there any reason why this is not a react component?

d2fong: class MetadataTip?

hmtinc: @maxkfranz  @d2fong  

The MetadataTip Class is just a regular class which manages tooltips, to help avoid random DOM elements remaining in the document and to make binding tooltips to cytoscape-popper easier. 

The initial intention was that since tippy.js takes in only HTML elements or a DOM selector, it would make more sense to use hyperscript to generate HTML elements and pass them into tippy. Then tippy would manage rendering elements and the MetadataTip class would ensure they didn't lurk around and take up too much space. 

In this case where each tooltip is short-lived and rendered outside the react tree by tippy.js anyway, is there a benefit to making the class MetadataTip a react component?

hmtinc:
@d2fong The metadatatip.js class got a bit bloated considering for what it was initially intended for, which was just to create a tippy tooltip. 

So, I have separated the generation code and array functions into separate files to make the the purpose of the functions clearer.

src/client/features/view/components/tooltips/generateContent.js
```diff
@@ -0,0 +1,218 @@
+const h = require('hyperscript');
+const config = require('../../config');
+
+//Generate HTML elements for a Parsed Metadata Field
+//Optional trim parameter indicates if the data presented should be trimmed to a reasonable length
+//Data Pair -> HTML
+function parseMetadata(pair, trim = true) {
```
d2fong: I think this function can be rewritten using the command pattern.

http://gameprogrammingpatterns.com/command.html

If I understand this function correctly, a pair can have one of many key types, and depending on the key you do different formatting right?

I think it would be better if you have a map for each key type and key type handler.

e.g
```js
const metaDataKeyMap = new Map()
.set('Standard Name', standardNameHandler)
...
const standardNameHandler = (pair) => {
   return makeTooltipItem(...);
}
```

and then parseMetaData can just be:
```js
const parseMetadata = (pair, trim = true) => {
  return metaDataKeyMap.get(pair[0])(pair);
}
```

src/client/features/view/components/tooltips/formatArray.js
```diff
@@ -0,0 +1,24 @@
+  //Push a metadata field to the bottom of the list
+  function pushToBottom(data, field) {
```
d2fong: I have a suspicion that this can be simplified with lodash.

Can you read over the lodash documentation to see if you can find a function that can do something similar?

src/client/features/view/components/tooltips/formatArray.js
```diff
@@ -0,0 +1,24 @@
...
+    for (var x in data) {
+      data[x][0] == field ? data.push(data.splice(x, 1)[0]) : 0;
+    }
+    return data;
+  }
+
+  //Order a given metadata data array
+  function orderArray(data) {
```
d2fong: Same thing for this function.

src/client/features/view/components/tooltips/generateContent.js
hmtinc: @d2fong  This does seem alot more cleaner to understand/read. 
I have converted all the if blocks into their own individual functions and set them in a map.

d2fong:
Also why was public/icon.png changed?

hmtinc:
@d2fong  The icon has a white line that went through the Icon, due to a alignment error. I fixed to SVG and PNG to not have this annoying line. 
screen shot 2017-11-10 at 1 20 07 pm

Also, all the recommended changes have been applied.

hmtinc:
@d2fong updated ReadMe to address issue #160

src/client/features/view/components/tooltips/generateContent.js
d2fong: I see a number of potential benefits of this approach:

1. It makes the reader aware that you do different things based on the meta data key; something that is harder to see in a cascading if statement.
2. It segments the code into more digestible chunks.
3. Makes each chunk more malleable and receptive to change.
4. It is easy to find via function name. E.g if you want to change anything to the old version of the function you have to scroll and look for a key string specifically. In the new version you can just look for the key in the function name itself.

src/client/features/view/components/tooltips/formatArray.js
```diff
@@ -0,0 +1,20 @@
...
+
+//Push a metadata field to the bottom of the list
+function pushToBottom(data, field) {
+  let withoutField = _.filter(data, pair => pair[0] !== field);
+  let withField = _.filter(data, pair => pair[0] === field);
+  return _.concat(withoutField, withField);
+}
+
+//Order a given metadata data array
+function orderArray(data) {
```
d2fong: This function still doesn't make sense

src/client/features/view/components/tooltips/generateContent.js
```diff
@@ -0,0 +1,238 @@
...
+}
+
+//Sort Database ID's by database name
+//Requires a valid database ID array
+//Array -> Array 
+function sortByDatabaseId(dbArray) {
+  let databases = {};
+
+  //Group Ids by database
+  for (let i = 0; i < dbArray.length; i++) {
```
d2fong: https://lodash.com/docs/4.17.4#groupBy

src/client/features/view/components/tooltips/generateContent.js
```diff
@@ -0,0 +1,238 @@
...
+  }
+}
+
+//Convert a generic array or string to an html list
+//Array -> HTML 
+function makeList(items, ulClass = 'ul.value-list', liClass = 'li.value-list-item') {
+  //Delete duplicates
+  items = deleteDuplicatesWithoutCase(items);
+
+  //Resolve possible errors
```
d2fong: I think all of this logic should be delegated to whoever calls makeList.

I.e create a strict contract for makeList.  Either make items be a a string or array.  Don't try to accommodate all of these different scenarios.

src/client/features/view/components/tooltips/generateContent.js
```diff
@@ -0,0 +1,238 @@
...
+    let databaseName = dbArray[i][0];
+    if (!databases[databaseName]) {
+      databases[databaseName] = [];
+    }
+    databases[databaseName].push(dbArray[i][1]);
+  }
+
+  //Produce final array with group names as the keys
+  dbArray = [];
+  for (let databaseName in databases) {
```
d2fong:
```js
dbArray = databases.map(db => { 
    ...
});
```

src/client/features/view/components/tooltips/generateContent.js
```diff
@@ -0,0 +1,238 @@
+const h = require('hyperscript');
+const config = require('../../config');
+
+
+//Handle name related metadata fields
+const standardNameHandler = (pair) => makeTooltipItem(pair[1], 'Approved Name: ');
+const displayNameHandler = (pair) => makeTooltipItem(pair[1], 'Display Name: ');
+const nameHandler = (pair, trim) => {
```
d2fong: The trim parameter is giving me a hunch that this function is doing too much.

src/client/features/view/components/tooltips/generateContent.js
```diff
@@ -0,0 +1,238 @@
...
+//Handle database related fields
+const dataSourceHandler = (pair) => {
+  let source = pair[1].replace('http://pathwaycommons.org/pc2/', '');
+  let link = generateDataSourceLink(source, 'Data Source: ');
+  return h('div.fake-paragraph', link);
+};
+const databaseHandler = (pair, trim) => {
+  //Sort the array by database names
+  let sortedArray = sortByDatabaseId(pair[1]);
+  if (sortedArray.length < 1) { return; }
```
d2fong: This function should always return html, it makes it hard to use this function if it doesn't consistently return something.

src/client/features/view/components/tooltips/generateContent.js
```diff
@@ -0,0 +1,238 @@
...
+  return h('div.fake-paragraph',
+    [
+      h('div.field-name', 'Database References:'),
+      h('div.wrap-text', h('ul.db-list', sortedArray.map(item => generateIdList(item, trim), this)))
+    ]);
+};
+
+//Handle type related fields
+const typeHandler = (pair, trim, key) => {
+  if (!(trim)) {
```
d2fong: Same problems here.  Don't use a trim flag here and always ensure you return html.

src/client/features/view/components/tooltips/generateContent.js
```diff
@@ -0,0 +1,238 @@
...
+    return h('li.db-item', generateDBLink(name, list[0], true));
+  }
+  else {
+    return h('li.db-item', h('div.db-name', name + ": "), list.map(data => generateDBLink(name, data, false), this));
+  }
+}
+
+//Generate a database link
+//Strings -> HTML 
+function generateDBLink(dbName, dbId, printId) {
```
d2fong: printId seems like some magic value to me. I can't understand what this is for.

src/client/features/view/components/tooltips/generateContent.js
hmtinc: Used _.GroupBy to sort the  values and then _.forEach to format the output

src/client/features/view/components/tooltips/generateContent.js
hmtinc: Removed logic from delete duplicates and created a toHTML function which maps any given value to the correct toHTML function.

src/client/features/view/components/tooltips/generateContent.js
hmtinc: The formatted array is now produced when the array is grouped, so there is no need to iterate twice.

src/client/features/view/components/tooltips/generateContent.js
hmtinc: Removed trim parameter from all handler functions and created separate functions to handle cases involving trim

src/client/features/view/components/tooltips/generateContent.js
hmtinc: All handler functions now always return an HTML element

src/client/features/view/components/tooltips/generateContent.js
hmtinc: Removed trim parameter from all handler functions and created separate functions to handle cases involving trimmed values.

src/client/features/view/components/tooltips/generateContent.js
hmtinc: `printId` is actually a boolean and is supposed to indicate if the databaseId is displayed on screen. 

I have renamed the variable to `isIdVisible` to make the purpose and type clearer

src/client/features/view/components/tooltips/formatArray.js
hmtinc: The purpose was to push a collection of elements to the bottom of a list. 
The name wasn't really that clear and the function wasn't really generic, so I have renamed it and make it slightly more generic.

src/client/features/view/components/tooltips/formatArray.js
```diff
@@ -0,0 +1,21 @@
...
+  let withoutField = _.filter(data, pair => pair[0] !== field);
+  let withField = _.filter(data, pair => pair[0] === field);
+  return _.concat(withoutField, withField);
+}
+
+//Push a collection of elements to the bottom of the list
+//Takes in a list of id's and pushes all matching items to the bottom of the list
+//Requires a valid list of ids
+function collectionToBottom(data, idList) {
+  idList.forEach(field => data = pushToBottom(data, field));
```
d2fong: Sorry I should have been more clear myself, but what I particularly don't understand is the mutation of data in that forEach line.

hmtinc: forEach simply just iterates over the IdList which just contains a collection of keys that need to be pushed to the bottom. 

Upon each foreach call, pushToBottom is called which returns a whole new ordered array. Since data contains the original unordered array, data is modified to ensure for the next call to pushToBottom has the correct partially sorted list.

hmtinc:
@d2fong All requested changes have now been applied.

src/client/features/view/components/tooltips/formatArray.js
d2fong: So is something like this the same?

```js
const collectionToBottom = (data, criteriaList) => {

   const matched = data.filter(pair => criteriaList.includes(pair[0]));
   const notMatched = _.difference(data, matched);
   
   return notMatched.concat(matched);
};
```

hmtinc: @d2fong That snippet works for pushing elements to the, but does not maintain order. The desired behaviour is that elements are pushed down in the order they appear in the criterial list.  Order is important is the list order defines what appears first in the tooltip. 

For example if the criteriaList was [obj2, obj1] and the data was [obj1, obj2, obj3], the desired result would be [obj3, obj2, obj1], but the given snippet would return [obj3, obj1, obj2] .

hmtinc: @d2fong  Your method works with the desired result if 2 additional lines were to be added
```js 
function collectionToBottom(data, criteriaList) {
  let matched = data.filter(pair => criteriaList.includes(pair[0]));
  let notMatched = _.difference(data, matched);
  let matchedGrouped =  _.groupBy(matched, 0);
  let matchedSorted = _.map(criteriaList, index => matchedGrouped[index].shift());
  return notMatched.concat(matchedSorted);
}
```
Although, I'm not sure how much cleaner that is compared to the existing.

d2fong: What about this:
```js
const collectionToBottom = (data, criteriaList) => {

   const matched = data.filter(pair => criteriaList.includes(pair[0]));
   const notMatched = _.difference(data, matched);
   
   return notMatched.concat(_.sortBy(matched, pair => pair[0]);
};
```

d2fong: But I guess you want a specific order still even if they are grouped by criteria list.

e.g if you want C, A, B.  You want all C items to come first, and then all A items, etc,  then the above still doesnt work?

hmtinc: Wouldn't that sort the matched list in lexicographic order? The desired order may not always alphabetical, as criterialList can be set to anything.

d2fong:
```js
const collectionToBottom = (data, criteriaList) => {

   const matched = data.filter(pair => criteriaList.includes(pair[0]));
   const notMatched = _.difference(data, matched);
   
   return notMatched.concat(_.sortBy(matched, pair => criteriaList.indexOf(pair[0]));
};
```

d2fong: But do you use the component function pushToBottom in other places too?

hmtinc: Didn't know you could user _.sortBy like that, but yeah that works and is a lot shorter/simpler that the 4 line solution I had. 

I have updated the function to follow the same pattern/style.

hmtinc: @d2fong pushToBottom was only used for collectionToBottom. The current updated version of this PR no longer has pushToBottom.

PR #154, opened 2017-11-08
Latest Version Fix & Connect cleanup

gaelder:
- Properly implemented determination of version when the api receives 'latest' as the version request
- Reworked connect to be in a single location and to cache a connection promise rather than create endless ones


src/server/database/utilities.js
```diff
@@ -2,6 +2,16 @@
 const r = require('rethinkdb');
 const config = require('./config');
 
+let connectionPromise = null;
+
+function connect() {
```
d2fong: @maxkfranz Do you think it makes sense to use lodash.memoize here?

d2fong: This is about caching the connection to rethinkdb.

maxkfranz: Yes, that would work

d2fong: @gaelder learn about memoize https://lodash.com/docs/4.17.4#memoize

And use that to cache the rethinkdb connection.

gaelder: @d2fong Done. Let me know if I did something weird.

PR #155, opened 2017-11-08
Added vast majority of Gary's suggested visual changes

jonahdlin:
In this PR I modified a lot of visual attributes as per Gary's suggestions including

- New layout dropdown menu
  - I could have used a dependency for this but none of the well-documented and standalone ones I could find would do what I wanted
  - Now displays information about each layout in the dropdown menu
- Redid download cards in the sidebar
  - The entire card is now clickable. To make sure this is clear to the user, it scales a little on hover
  - Added react-loader loading icon to downloads (I couldn't figure out how to have it work properly with the png download though)
- Removed cola.js layout option, since it's very similar to CoSE-Bilkent and it may cause confusion to have two force-directed layouts
- Source name in info menu now links to datasource pages
- Small wording changes

As is the case for a lot of my PRs nowadays the code isn't too pretty, but we want to get a working demo with all of Jeff and Gary's aesthetic suggestions implemented so we can send it out to the whole Pathway Commons Dev team.


src/client/features/view/components/menu/index.js
```diff
@@ -3,6 +3,11 @@ const h = require('react-hyperscript');
 const Link = require('react-router-dom').Link;
 const classNames = require('classnames');
 
+const layoutConf = require('../../../../common/cy/layout');
+
+const Dropdown = require('../../../../common/dropdown').Dropdown;
```
d2fong: Pretty sure you can do this:
```js
const { Dropdown, DropdownOption } = require('../../../../common/dropdown');
```

src/client/features/view/components/sidebar/menus/components/downloadOption.js
```diff
@@ -23,7 +24,11 @@ class DownloadOption extends React.Component {
     };
   }
 
-  handleDownloadClick(type) {
+  handleDownloadClick(evt, type) {
+    // Don't do anything when a link is clicked. Should not download just because
+    // user clicks a link
+    if (evt.target.tagName === 'A') return;
```
d2fong: Try:
```js
evt.preventDefault();
```
Instead of checking for the tagName and see if it works.

src/styles/features/view/animations/loadingSpin.css
```diff
@@ -0,0 +1,9 @@
+@keyframes loading-spin {
```
d2fong: Is this loading spinner specifically for the .png download that react-loader didn't work with?

Can you provide more detail about how react-loader doesn't work with the .png download?

src/styles/index.css
```diff
@@ -11,3 +11,4 @@
 @import "./features/paint.css";
 @import "./features/view.css";
 @import "./features/admin.css";
+@import "./common/index.css";
```
d2fong: common css should be after app.css but before features

src/styles/vars.css
```diff
@@ -4,11 +4,13 @@
...
   --cloud: #ECF0F1;
   --cloud-dark: #d3d8da;
+  --cloud-darker: #b5b9bb;
 
   /* Colours used by other modules */
   --accent-1-colour: var(--green-sea);
   --light-base-colour: var(--cloud);
   --light-base-colour-dark: var(--cloud-dark);
+  --light-base-colour-darker: var(--cloud-darker);
   --dark-base-colour: var(--midnight-blue);
```
d2fong: @maxkfranz is this what you had in mind for semantic css var names?  I find it more difficult to remember ```accent-1-colour``` vs ```green-sea```.

src/styles/features/view/animations/loadingSpin.css
jonahdlin: The loading spinner CSS I left in by mistake before I realized react-loader would work fine.

As for the .png download, unlike the other formats, comes straight from cytoscape.js, and is then saved with saveAs from file-saver. I couldn't figure out how to tell when the procedure is done since neither saveAs nor cy.png() return promises or take callbacks.

src/client/features/view/components/menu/index.js
jonahdlin: Yup, that works. I've made that change and I'll keep in mind to change others when I come across them.

src/styles/features/view/animations/loadingSpin.css
d2fong: Have you tried something like this already?
```js
this.setState({loading: true});
cy.png(...);
saveAs(blob, ..., ...);
this.setState({loading: false});
```

d2fong: You can also do:

```js
const saveImage = new Promise((resolve, reject) => {
   const img = cy.png();
   saveAs(img, ...., ...);
   resolve(null);
});

saveImage.then(() => this.setState({loading: false});

```

d2fong: It might be that saveAs the problem

src/client/features/view/components/sidebar/menus/components/downloadOption.js
jonahdlin: `evt.preventDefault()` doesn't work in this case it would seem because it tries to prevent the default behaviour of the event. In this case I want the default behaviour and not the behaviour I define, so if a user clicks a link inside the download option, the link opens, but the download does not happen.

d2fong: So when a user clicks an 'a' tag, you want to open the link and download a file?

jonahdlin: No, I want to open the link and not download the file.

d2fong: Do you always want to open the link and not download the file or does that depend on the type of file you are downloading?

jonahdlin: Doesn't matter the file type. The links are clickable without downloading the file. The whole card is clickable for the download except the links.

src/styles/features/view/animations/loadingSpin.css
jonahdlin: I'll give that a shot tomorrow with some pathways.

jonahdlin: That code runs into the same problems as the previous code. I've done more testing on it and it seems to be a synchronous process, that for some reason holds the entire page from rendering while executing. If we do
```js
this.setState({loading: true});
cy.png(...);
saveAs(blob, ..., ...);
this.setState({loading: false});
```
then the loader is correctly set to loading before the process executes, and stops loading afterwards, but never displays while it's active.

@hmtinc suggested wrapping the process in a `setTimeout`. This works as long as it's called as a callback to the first `setState`, i.e.
```js
this.setState({ loading: true }, () => {
  setTimeout(() => {
    saveAs(this.props.cy.png({
      ...
    }), ...);
    this.setState({ loading: false });
  }, 1);
});
```
Doesn't matter how long the timeout is (1ms works fine), and it has to be wrapped in the callback for `this.setState`.

src/client/services/cdc/index.js
```diff
@@ -1,11 +1,21 @@
 const io = require('socket.io-client');
-let socket = io.connect('/');
+let socket = io.connect('192.168.81.233:3000');
```
d2fong: Revert this change

Issue #159, opened 2017-11-10
In src/server/index.js only use one routes object


d2fong:
in ```server/index.js```, the code currently looks like this:
```js
app.use('/', require('./routes/databaseRoutes/sockets')(io));
app.use('/layout', require('./routes/databaseRoutes/rest'));
app.use('/', require('./routes/'));
```

Can you try changing it so that it looks like:

```js
app.use('/', require('./routes'));
```

And in ```server/routes/index.js```, figure out a way to combine them into one module.export

Maybe try something like:
```js
router.get('/layout', require('./routes/databaseRoutes/rest'));
```

Not sure if it will work or not, but I think it is important to only expose a singular module.export via the index file of the routes folder. 

Please try to find a way to combine the routes into one export in the routes/index.js file.

PR #163, opened 2017-11-10
Endpoint rework

gaelder:
- Renamed endpoints to be consistent and better reflect their function
- Refactored route controller to better separate getting graph from layout.
- Refactored router creation to simplify the structuring of server/index.js (#159)


src/server/index.js
```diff
@@ -46,9 +48,8 @@ app.use(morgan('dev', {
 app.use(cookieParser());
 app.use(express.static(path.join(__dirname, '../..', 'public')));
 
-app.use('/', require('./routes/databaseRoutes/sockets')(io));
-app.use('/layout', require('./routes/databaseRoutes/rest'));
 app.use('/', require('./routes/'));
+defineSockets(io);
```
d2fong: Can socket routes go into routes too?

src/server/database/update.js
```diff
@@ -53,11 +54,24 @@ function saveGraph(pcID, graphID, releaseID, existingGraph, newGraph,connection)
 function updateGraph(pcID, releaseID, cyJson, connection, callback) {
   let graphID = uuid();
 
+  let nodeSort = (n1, n2) => {
```
d2fong: So node ids are numbers?

You can also use lodash to sort by key I believe.

src/server/index.js
d2fong: The reason that this is important is because it encapsulates all the routes code in one place. You also want to minimize the point of contact between routes and the server index file.

d2fong: You also want to minimize contact with server/index in general. Most of it was automatically generated and serves as glueing config stuff together.

d2fong: Looking into this more, it seems that it does make some sense to keep defineSockets in the server file.

src/server/database/update.js
gaelder: Node id's are uuid strings

src/server/index.js
gaelder: To clarify, what changes should I make here?

src/server/database/update.js
maxkfranz: `_.sortBy( array, val => val.comparableValue )`

src/server/index.js
maxkfranz: Use modules

- [ ] Make /server/io.js
- [ ] Put the `io` creation there and `module.exports = io`, i.e. `io` is a singleton
- [ ] server/index requires server/io.js
- [ ] the separate files that handle different socket.io events require server/io.js as well

src/server/index.js
```diff
@@ -49,7 +46,10 @@ app.use(cookieParser());
 app.use(express.static(path.join(__dirname, '../..', 'public')));
 
 app.use('/', require('./routes/'));
-defineSockets(io);
+app.use('/api', require('./routes/rest'));
+require('./io').set(server);
+require('./routes/sockets');
```
gaelder: @maxkfranz I made the changes to the best of my understanding, but this still feels like something isn't quite right. I needed to add the second require statement so that the sockets were actually defined in the io object.

src/server/index.js
gaelder: See changes below. I've tagged a spot that still seems iffy.

jvwong:
Resolves dependencies in fab450e150620d081dacf1a12df15131f24d983c

PR #172, opened 2017-11-13
Replaced single socket.io use with fetch

jonahdlin:
Also added infrastructure for eventually doing live updates. More has to be fleshed out for this to work.

### DIFF
--- a/src/client/services/cdc/index.js
+++ b/src/client/services/cdc/index.js
@@ -15,7 +15,6 @@ const CDC = {
   // Send a diff in a node to the backend. The backend will deal with merging these diffs into
   // a layout
   submitNodeChange(uri, version, nodeId, bbox) {
-    console.log('SENT A NODE CHANGE');
     socket.emit('submitDiff', {
       uri: uri,
       version: version.toString(),
@@ -27,7 +26,6 @@ const CDC = {
   },
 
   submitLayoutChange(uri, version, layout) {
-    console.log('SENT A LAYOUT CHANGE');
     socket.emit('submitLayout', {
       uri: uri,
       version: version,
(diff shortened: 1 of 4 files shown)


src/client/services/cdc/index.js
```diff
@@ -2,26 +2,14 @@ const io = require('socket.io-client');
...
-        updateFunction(cyJSON);
-      } catch(err) {
-        if (msg.length > 0) {
-          console.log(JSON.parse(msg));
-          //console.log(msg);
-        } else {
-          throw err;
-        }
+  getGraphAndLayout(uri, version) {
+    return fetch(`http://${window.location.hostname}/get-graph-and-layout?uri=${encodeURIComponent(uri)}&version=${version}`, {
```
d2fong: Does the following code not work?
```js
fetch(`/get-graph-and-layout?uri=${encodeURIComponent(uri)}&version=${version}`, ... );
```

d2fong: You should not be using window at all when possible unless you like programming with global variables.

src/client/services/cdc/index.js
```diff
@@ -2,26 +2,14 @@ const io = require('socket.io-client');
...
+        'Accept': 'application/json'
       }
-    });
-  },
-
-  // Request a graph from the server. Should send back a cyJSON graph
-  requestGraph(uri, version) {
-    socket.emit('getLayout', {uri: uri, version: version.toString()});
+    }).then(res => res.json())
+    .then(data => JSON.parse(data));
```
d2fong: Redundant code.  res.json() already parses the response body.

https://developer.mozilla.org/en-US/docs/Web/API/Body/json

src/client/services/cdc/index.js
jonahdlin: I'll change it to that. It works. Thanks for the tip!

src/client/services/cdc/index.js
jonahdlin: Resolved. Turns out we had left in a stringify on the server-side, which is why we still needed that on the frontend.

PR #181, opened 2017-11-15
Adjust api nomenclature so backend and front end work together again

gaelder:


gaelder:
@jvwong This should fix master. Please let me know if that is not the case ASAP

Issue #178, opened 2017-11-15
Entry page visual overhaul


jonahdlin:
Entry page should:

1. Indicate Pathway Commons
2. Indicate app function
3. Example search queries
4. Describe search contents

Inspiration : https://www.ebi.ac.uk/QuickGO/

d2fong:
Based on initial feedback it seems that people want it changed to look like a hybrid of your inspiration page and the previous search page:
![screen shot 2017-11-16 at 12 17 27 pm](https://user-images.githubusercontent.com/2328291/32905507-2639f70e-cac8-11e7-8dec-a136f4b9104a.png)

PR #182, opened 2017-11-15
Entry page changed to be more descriptive, and is themed like search

jonahdlin:
Changed the entry page to look more like the search page, and included all points from #178.

The big "Search" is a link to the search page, as are all the suggestions below the search bar (they link to a search for that term). I picked three pretty arbitrary ones as examples, so suggestions for what to provide as examples are definitely welcome.

screen shot 2017-11-15 at 11 11 20 am

### DIFF
--- a/src/client/features/entry/index.js
+++ b/src/client/features/entry/index.js
@@ -1,80 +1,93 @@
 const React = require('react');
 const h = require('react-hyperscript');
+const { Link } = require('react-router-dom');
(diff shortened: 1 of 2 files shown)


d2fong:
@jvwong can you confirm that this is what you want it to look like?

d2fong:
Get explicit approval from 3 of Gary, Jeff, Igor, Max and then I will merge it.

Issue #83, opened 2017-10-31
Add node search capability


hmtinc:
@cannin 
https://github.com/PathwayCommons/pathways-search/issues/130

>This is a TP53 pathway (title).
>http://www.pathwaycommons.org/pathways/#/view?uri=http%3A%2F%2Fidentifiers.org%2Freactome%2FR-HSA-6803211
>It isn't intuitive where TP53 is. Found it by uncollapsing the nodes.

PR #186, opened 2017-11-16
Added basic node search functionality 

hmtinc:
#83 - Basic node search has been implemented allow for users to search the view for nodes. The search currently searches based on the label and will search all nodes including nested nodes. 

#### Basic Search
Basic search will search all nodes for partial matches and will ignore case. To conduct a basic search, simply type a keyword in the search bar.  

 - e.g `tp53`, `bid`, `BA`, `BAX` 

#### Exact Search
Exact search will search all nodes for exact matches, but will ignore case as well. To conduct an exact search, simply prefix the search query with `exact:` 
 
 
 - e.g `exact:tp53`, `exact:BaX`, `exact:ADP`
 
#### Regular Expression Search 
Regular expression search is an additional method that has been included to allow users to define more complicated search queries.  To conduct a regular expression based search, simply prefix the search query with `regex:`
 
 - e.g `regex:/tp53/i`, `regex:/BAX/ `

### DIFF
--- a/src/client/features/view/components/menu/index.js
+++ b/src/client/features/view/components/menu/index.js
@@ -68,7 +68,7 @@ class Menu extends React.Component {
             h('h4', `${this.props.name} | ${this.props.datasource}`)
           ]),
           h('div.search-nodes', {
-            onChange : query => searchNodes(query, this.props.cy),
+            onChange : e => searchNodes(e.target.value, this.props.cy),
             title: 'Search for Nodes'
           }, [
             h('div.view-search-bar', [h('input.view-search', {type : 'text', placeholder: 'Entity Search'})])
--- a/src/client/features/view/components/menu/index.js
+++ b/src/client/features/view/components/menu/index.js
@@ -70,7 +70,7 @@ class Menu extends React.Component {
             onChange : query => this.props.searchNodes(query),
             title: 'Search for Nodes'
           }, [
-            h('input', {type : 'text', placeholder: 'Search for Nodes'})
+            h('input', {type : 'text', placeholder: 'Entity Search'})
           ]),
           h('div.layout-dropdown-button', {
             onClick: () => this.setState({dropdownOpen: !this.state.dropdownOpen}),
--- a/src/client/features/view/components/menu/index.js
+++ b/src/client/features/view/components/menu/index.js
@@ -7,6 +7,8 @@ const layoutConf = require('../../../../common/cy/layout');
 
 const { Dropdown, DropdownOption } = require('../../../../common/dropdown');
 
+const searchNodes = require('./search');
+
 const tippy = require('tippy.js');
 
 /* Props
@@ -37,7 +39,6 @@ class Menu extends React.Component {
       touchHold: true
     });
   }
-
   render() {
     const layoutItems = this.props.layouts.map((layout, index) => {
       return (
@@ -67,7 +68,7 @@ class Menu extends React.Component {
             h('h4', `${this.props.name} | ${this.props.datasource}`)
           ]),
           h('div.search-nodes', {
-            onChange : query => this.props.searchNodes(query),
+            onChange : query => searchNodes(query, this.props.cy),
             title: 'Search for Nodes'
           }, [
             h('input', {type : 'text', placeholder: 'Entity Search'})
--- a/src/client/features/view/components/menu/index.js
+++ b/src/client/features/view/components/menu/index.js
@@ -66,6 +66,12 @@ class Menu extends React.Component {
           h('div.title-container', [
             h('h4', `${this.props.name} | ${this.props.datasource}`)
           ]),
+          h('div.search-nodes', {
+            onChange : query => this.props.searchNodes(query),
+            title: 'Search for Nodes'
+          }, [
+            h('input', {type : 'text', placeholder: 'Search for Nodes'})
+          ]),
           h('div.layout-dropdown-button', {
             onClick: () => this.setState({dropdownOpen: !this.state.dropdownOpen}),
             title: 'Rearrange the entities on screen'
--- a/src/client/features/view/components/menu/index.js
+++ b/src/client/features/view/components/menu/index.js
@@ -71,7 +71,7 @@ class Menu extends React.Component {
             onChange : query => searchNodes(query, this.props.cy),
             title: 'Search for Nodes'
           }, [
-            h('input.view-search', {type : 'text', placeholder: 'Entity Search'})
+            h('div.view-search-bar', [h('input.view-search', {type : 'text', placeholder: 'Entity Search'})])
           ]),
           h('div.layout-dropdown-button', {
             onClick: () => this.setState({dropdownOpen: !this.state.dropdownOpen}),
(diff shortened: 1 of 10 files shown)


src/client/features/view/components/menu/search.js
```diff
@@ -0,0 +1,96 @@
+const _ = require('lodash');
+
+//Apply hover styling to a collection of nodes
+function updateStyling(style, matched, cy) {
+  _.forEach(matched, node => {
```
d2fong: Is the ```matched``` parameter a cy.js collection?

If so, you should rename ```matched``` to ```matchedNodes```
and then use ```nodes.forEach``` instead of ```lodash.forEach```

i.e

```js
  matchedNodes.forEach(node => {
```

src/client/features/view/cy/events/click.js
```diff
@@ -9,6 +9,9 @@ const bindClick = (cy, callback) => {
       let name = data.label;
       let cy = evt.cy;
 
+      console.log(evt.target.scratch('_search-style-before'));
```
d2fong: Delete log statements

src/client/features/view/components/menu/search.js
```diff
@@ -0,0 +1,96 @@
...
+    let isCompartment = node.data('class') === 'compartment';
+
+    //Collapse Expanded Targets
+    if (node.isExpanded() && !isCompartment) { node.collapse(); }
+
+    //Apply style to the parent
+    if (node.isChild() && node.parent().data('class') !== 'compartment') { node = node.parent(); }
+
+    applySearchStyle(cy, node, style);
```
d2fong: Remove cy from the applySearchStyle params

src/client/features/view/components/menu/search.js
hmtinc: Yes, it is a collection. I have renamed it and used regular forEach().

src/client/features/view/components/menu/search.js
hmtinc: Cy has been removed

src/client/features/view/cy/events/click.js
hmtinc: Didn't realize that was still there. I have removed it in the last commit.

hmtinc:
@d2fong  The requested changes have been made

src/client/features/view/components/menu/search.js
```diff
@@ -0,0 +1,96 @@
...
+}
+
+//Search for nodes that match an entered query
+function searchNodes(query, cy) {
+  const searchValue = query.target.value;
+  const isBlank = _.isString(searchValue) ? !!_.trim(searchValue) : false;
+  const isRegularExp = _.startsWith(searchValue, 'regex:') && validateRegex(searchValue.substring(6));
+  const isExact = _.startsWith(searchValue, 'exact:');
+
+  let visibleNodes = Array.prototype.slice.call(cy.nodes(), 0);
```
d2fong: What does this line do?

src/client/features/view/components/menu/search.js
```diff
@@ -0,0 +1,96 @@
...
+
+  //Add children of nodes to nodes list
+  const allChildNodes = visibleNodes.map(node => node.data('compoundCollapse.collapsedCollection'));
+  const validChildNodes = _.compact(_.flattenDeep(allChildNodes.map(collection => collection ? Array.prototype.slice.call(collection, 0) : null)));
```
d2fong: What does this line do?

src/client/features/view/components/menu/search.js
```diff
@@ -0,0 +1,96 @@
...
+  const nodes = _.union(visibleNodes, validChildNodes);
```
d2fong: Generally for cytoscape collections I would prefer you use the cytoscape versions of these functions.

http://js.cytoscape.org/#eles.slice

http://js.cytoscape.org/#collection/building--filtering

src/client/features/view/components/menu/search.js
```diff
@@ -0,0 +1,96 @@
...
+  //Remove all search styling
+  applySearchStyle(cy.nodes(), baseStyle);
+
+  //Apply styling
+  if (matched.length > 0 && isBlank) {
+    updateStyling(searchStyle, matched, cy);
+    cy.fit();
+  }
+  else {
+    cy.fit();
```
d2fong: should we be calling cy.fit when there are no matches?

src/client/features/view/components/menu/search.js
hmtinc: I have simplified line 47-50 into just 2 lines to make the code a lot easier to read and understand. This line is no longer needed.

src/client/features/view/components/menu/search.js
hmtinc: These lines took an array of collections and appended them to the main node list. This is no longer needed, as the last commit simplifies the steps .

src/client/features/view/components/menu/search.js
hmtinc: I have changed the implementation to use cytoscape eles.union()

src/client/features/view/components/menu/search.js
hmtinc: I guess that would be the better option. The initial idea was that if no results were found, the entire graph would be displayed to indicate nothing was highlighted. 

I have removed it for now, maybe in a future revision there can be a `no results found` message.

src/client/features/view/components/menu/index.js
```diff
@@ -66,6 +67,12 @@ class Menu extends React.Component {
           h('div.title-container', [
             h('h4', `${this.props.name} | ${this.props.datasource}`)
           ]),
+          h('div.search-nodes', {
+            onChange : query => searchNodes(query, this.props.cy),
```
d2fong: There is a slight mismatch in naming here, query in this case should be called event or e for short, which is consistent with the react docs.  https://reactjs.org/docs/handling-events.html

You could change the function signature for searchNodes to actually take in the string value for the query.

i.e
```js
  onChange: e => searchNodes(e.target.value, this.props.cy);
```

This reduces coupling between the event and searchNodes because searchNodes doesn't have to assume it is getting a event with target.value, all it has to assume is that query is a string.

src/client/features/view/components/menu/search.js
```diff
@@ -0,0 +1,94 @@
...
+  }
+}
+
+function applySearchStyle(eles, style) {
+  eles.style(style);
+}
+
+//Search for nodes that match an entered query
+function searchNodes(query, cy) {
+  const searchValue = query.target.value;
```
d2fong: See above.

src/client/features/view/cy/events/hover.js
```diff
@@ -174,4 +174,7 @@ const bindHover = (cy) => {
 
 };
 
-module.exports = bindHover;
\ No newline at end of file
+module.exports = {
+  bindHover,
+  storeStyle
```
d2fong: Do you still need storeStyle?

src/client/features/view/components/menu/index.js
hmtinc: I have changed the function signature to reduce coupling

src/client/features/view/components/menu/search.js
hmtinc: Removed, as query is already the value in the newer commit

src/client/features/view/cy/events/hover.js
hmtinc: No, storeStyle should have been removed form module.exports. It has now been removed.

hmtinc:
@d2fong  All additional changes have now been implemented.

PR #197, opened 2017-11-17
Search scaling and visuals

jonahdlin:
Changed a few visuals with the search page including
- Changed branding on left to include "Search"
- Centered results and search bar like in old, old search
- Added clickable examples underneath search bar
- Made it more easily accessible on mobile
  - On smaller screens, the branding and images disappear
  - The search results contract so they don't go offscreen
  - The search bar also contracts and tabs are scrollable on small widths

@jvwong what do you think?

New search:
screen shot 2017-11-17 at 11 51 07 am

Old search:
screen shot 2017-11-17 at 11 50 49 am

For Dylan/Max, I wanted the search branding to be off to the left, but the search bar to be centered. I absolutely positioned the branding div, which means that when the screen gets smaller, it overlaps the search bar. Without hard-coding values, the best solution I could come up with is to use react refs and check in a resize listener if the two are overlapping with

```js
const searchBarLeft = this.searchBar.getBoundingClientRect().left;
const searchBrandingRight = this.searchBranding.getBoundingClientRect().right;
const isOverlapping = searchBarLeft <= searchBrandingRight;
```

Can you guys think of a pure CSS solution?


d2fong:
What does it look like on mobile?

jonahdlin:
Mobile view. Tabs are scrollable left and right.

screen shot 2017-11-17 at 12 39 23 pm

d2fong:
Also @jvwong @jonahdlin do you think we should make the search consistent with the view in terms of adding black borders to things?

e.g
screen shot 2017-11-17 at 12 42 33 pm
screen shot 2017-11-17 at 12 42 28 pm

Maybe add the same border style to the search ui.

d2fong:
@jonahdlin have you looked into using media queries?
https://developer.mozilla.org/en-US/docs/Web/CSS/Media_Queries/Using_media_queries

jonahdlin:
I've been using media queries in the view. I'm not sure which I should be using.

jonahdlin:
For the border thing, I prefer the look of the search to that of the view. I'm going to experiment with a dark theme like in Search for the view to see what I can come up with.

maxkfranz:
I would prefer to remove borders where possible to make things consistent rather than adding more of them.

maxkfranz:
You should be using flex positioning to get the alignment how you want.  But frankly, I think it's better to keep the UI to the left side anyway.  That leaves open the possibility of adding some UI to the right side in future (like a graph preview on hover).

jonahdlin:
New light theme for search, to be more consistent with view. I moved it back to the left side and added some view-like hover styling to the tabs.

screen shot 2017-11-20 at 10 33 26 am

screen shot 2017-11-20 at 10 39 59 am

I also changed all the scaling to `@media` queries, and changed all `@media` queries across the app to be variable. The size of a "small" screen can now be changed in vars.css. To do that, I added a dev dependency postcss-custom-media.

src/client/features/search/index.js
```diff
@@ -95,22 +96,42 @@ class Search extends React.Component {
...
+      const newQueryState = _.assign({}, state.query);
+      newQueryState.q = search;
+      return h('span.search-example', { onClick: () => this.setAndSubmitSearchQuery(search) }, search);
+    });
+
+    let i = 1;
+    while (i < examples.length) {
+      examples.splice(i, 0, ', ');
+      i += 2;
+    }
```
d2fong: what is this while loop supposed to do?

src/client/features/search/index.js
```diff
@@ -95,22 +96,42 @@ class Search extends React.Component {
     this.getSearchResult();
   }
 
+  setAndSubmitSearchQuery(searchTerm) {
```
d2fong: Why is this function needed?

src/client/features/view/config.js
```diff
@@ -23,9 +23,16 @@ const databases = [
   ['CAS', 'http://identifiers.org/cas/', '']
 ];
 
+// Upper limit for mobile sizing in pixels, this is currently based off the iphone 6 plus
+const mobileUpperLimit = {
```
d2fong: Why is this needed?

src/client/features/search/index.js
jonahdlin: It's used for the suggestions underneath the bar (e.g. TP53, Glycolysis, Ethanol). All those suggestions are clickable so I needed a way to set the current search query and submit it in one function.

d2fong: You mean like this function?

https://github.com/PathwayCommons/app-ui/pull/197/files#diff-a1f9f779cabf8780553a2f0a90378e43R76

jonahdlin: Yep, like that one. Didn't see it before. Just fixed it.

d2fong: You can just use:

```js
this.setQueryState({
  q: exampleText
});
```
It calls submitSearchQuery.

However, it isnt clear that ```setQueryState``` calls ```submitSearchQuery```, so you should keep the ```setQueryState``` method body, change the name to ```setAndSubmitSearchQuery``` and change every instance that ```setQueryState``` is used to ```setAndSubmitSearchQuery```.

The reason to use setQueryState instead of your version of the function is that it is more general.  E.g if you want to set query type or any other query field that function will handle that too.

jonahdlin: That makes sense. Better to keep things reusable. I've made the changes and committed them.

src/client/features/search/index.js
jonahdlin: The example searches underneath the search bar is produced from an array of suggested search queries in the render function. After the map happens, I want unclickable commas separating the clickable spans, which is what that while loop does. I'm sure there's a better way though.

The essential problem is taking an array and splicing a value between all entries,
`[1, 2, 3]` --> `[1, val, 2, val, 3]`

src/client/features/view/config.js
jonahdlin: It's old code from before I switched search to `@media` I forgot to remove. Removed it.

src/client/features/search/index.js
d2fong: (1) Make a function that returns the `h()` part for an example.

Basically you can create functions for simple React "components", e.g.

```js
let Foo = props => h('div.foo', { onBar: props.onBar });
```

(2) Add the examples manually, e.g.

```js
let examples = [
  h( Example, { q: 'p53' } ),
  h('span', ', '),
  h( Example, { q: 'pcna' } )
];
```

jonahdlin: Fixed.

Issue #213, opened 2017-11-20
Fix styling on database references 


hmtinc:
Database references and plain text values should be formatted such that they appear on the same line and do not break mid sentence.

Issue #189, opened 2017-11-16
Remove datasource from the tooltip 


hmtinc:
Datasource is usually the same as the pathway datasource, so it should be removed from the tooltip to avoid confusion.

Issue #187, opened 2017-11-16
Fix tooltip field ordering


hmtinc:
The tooltip expansion loads new fields at the top which causes the ordering to change. The ordering between views should remain consistent.

hmtinc:
@gbader  @jvwong 

I implemented some of the suggested changes and have included publication info in the tooltip. 

![screen shot 2017-11-20 at 10 38 23 am](https://user-images.githubusercontent.com/1313810/33026787-39d89be8-cddf-11e7-9977-ff207f8ca1f3.png)

![screen shot 2017-11-20 at 10 33 43 am](https://user-images.githubusercontent.com/1313810/33026790-3bbd5674-cddf-11e7-955b-95aa88cf21d2.png)

Issue #180, opened 2017-11-15
Publication info via ID


jonahdlin:
Using BioPAX metadata, we can get the title and abstract for publications from this service that allows you to fetch publication info via ID (https://eutils.ncbi.nlm.nih.gov/entrez/eutils/efetch.fcgi?db=pubmed&retmode=xml&rettype=abstract&id=), for nodes or for whole pathways. Would be neat to include abstracts and titles of publications somewhere in the view.

hmtinc:
@jvwong  Is PubMed the only source for publications?

jvwong:
From the BioPAX spec: "PublicationXref....Comment: Publication xrefs should make use of PubMed IDs wherever possible. The db property of an xref to an entry in PubMed should use the string “PubMed” and not “MEDLINE”."

If it wasn't indexed by Pubmed I wouldn't cite it.

hmtinc:
@jvwong @jonahdlin 
So, I have implemented basic parsing functionality for database id's into the Metadata Mapper. The graph-generation module will now cache the title of publications and authors of publications. 

Is there any other fields that might be needed for publications?

PR #216, opened 2017-11-20
Implemented tooltip and search suggestions

hmtinc:
### Tooltip 

- #187 -  Fixed tooltip ordering. All the tooltips will always maintain the order of the values. 
- #180 -  Replaced Publication Ids with publication names
 screen shot 2017-11-20 at 10 38 23 am

- #189 - Removed Datasource/Display Name from the tooltip
- #213 - Database References and field-values have been formatted as comma separated lists

### Search 
#212 - Set the Overlay Padding to 0


d2fong:
So you were able to get it working by using promises in the metadata generation?

hmtinc:
@d2fong Yeah, I got it to work by wrapping all the generation functions in a promise.

PR #237, opened 2017-11-22
PC Dev Suggestions

### LINKED ISSUE #190: Perhaps, hide "REPLACED ..." annotations (BioPAX comments)
E.g., if I right-click on an entity node (e.g., a Protein), then - "Additional info", I can see a comment like, e.g., 
`REPLACED http://www.reactome.org/biopax/61/48887#Protein16326`
which I'd rather hide/ignore, because 
- such comments were added mainly for debugging Pathway Commons data integration code;
- original physical _entity_ (e.g., in Reactome) URIs are often not standard, stable, nor resolvable to a web page...

### LINKED ISSUE #198: Zoom Is Adjusted Slightly During Search Typing on Pathway Page Highlighting
It looks like the zoom changes each time I type in a letter in the search to highlight a node. Can the zoom level be frozen?

### LINKED ISSUE #235: Delete Duplicate Cellular Locations
Some cellular locations have duplicate entries and should be filtered for unique values only.

### LINKED ISSUE #236: Metadata does not appear for molecular interactions
Tooltips on molecular interaction graphs do not appear. Metadata mapper quits early for each entity in a molecular interaction, which results in a null entry for parsedMetadata.

### LINKED ISSUE #242: Tooltips don't close if complex gets collapsed
If a tooltip is opened for a node within a complex:
image
then the complex is collapsed:
image
The tool tip does not disappear until the first click event after the complex is reopened
image

hmtinc:
### Zoom Issues
closes #198 - The padding of each overlay has been set to 0 and the graph is automatically fit after each layout run. This fixes all the jittering issues that were present in the beta 

### Tooltip  Issues
closes #190 - All comments that contain `REPLACED` have been filtered out.
closes #235 - Duplicate Cellular Locations has been removed from the tooltip 
closes #242 - Fixed a bug where a tooltip would not disappear when a complex is collapsed 

### Metadata Issues
closes #236 - Tooltips now appear on molecular interactions.


src/client/features/view/components/menu/index.js
```diff
@@ -51,6 +52,10 @@ class Menu extends React.Component {
       }
     });
     layout.run();
+    layout.pon('layoutstop').then(function( event ){
```
d2fong: I think this can be combined with the other layoutstop function 8 lines up.

d2fong:
Also it might make sense to promote this code to the common/cy module.  Can you see if you can refactor it to be less coupled from the view?

src/client/features/view/components/menu/index.js
hmtinc: It indeed can. The other one was just recently added by another PR. I have combined the two together

hmtinc:
@d2fong I have refactored the search code and moved search to/common/cy. 
The module takes in a cytoscape object `cy` and a search query string, and applies the search styling to correct nodes, so it should be able to used elsewhere in the project now.

PR #255, opened 2017-11-24
Port changed to 2000 for development

jonahdlin:
This way we can host both on the VM


maxkfranz:
You shouldn't change the source for this.  The default should always be 3000.  Set an environment variable in your VM to change the port of a particular instance.

jonahdlin:
My mistake. I didn't know that until after I had done it. I already set it to 2000 in Jenkins, and reverted the default here to 3000 in #265

PR #256, opened 2017-11-27
Update to beta version of uglify JS to support ES6 syntax

jonahdlin:
Currently the development build on Jenkins fails because `uglify-js` doesn't support some ES6 syntax. Updating the uglify version to the beta fixes the production build.


maxkfranz:
Babel should be running prior to uglify, so uglify should be running on ES5 code.

PR #268, opened 2017-11-29
Tooltip Revision 3 

### LINKED ISSUE #261: Add expansion buttons to Comments, Names, and Links
Instead of a singular view more info button, there should be a view more link on all the fields that go over the default entry limit.

hmtinc:
.This PR implements all the suggested tooltip UI changes made during the Tuesday meeting with Gary. 

### Tooltip Style Changes
closes #226 
- Removed type field from tooltip, but kept the type name
- Changed `approved name` to `name`
- Changed Database References` to Links
- Changed default position to `bottom`
- Assigned a max height to tooltip

### Tooltip Expansion Changes
closes #261

The tooltip expand button has been removed. Comments, Links, and Synonyms now have their options to individual expand and close data.

####  Regular tooltip
screen shot 2017-11-29 at 1 26 34 pm

#### Scrollable tooltip with max height 
screen shot 2017-11-29 at 1 20 46 pm

#### Expandable Synonymns
expand ex

#### Expandable Comments
expand com

### DIFF
--- a/src/client/common/tooltips/generateContent.js
+++ b/src/client/common/tooltips/generateContent.js
@@ -5,7 +5,7 @@ const config = require('./config');
 
 
 //Handle name related metadata fields
-const standardNameHandler = (pair) => makeTooltipItem(pair[1], 'Approved Name: ');
+const standardNameHandler = (pair) => makeTooltipItem(pair[1], 'Name: ');
 const standardNameHandlerTrim = (pair) => standardNameHandler(pair);
 
 const nameHandlerTrim = (pair) => {
@@ -351,7 +351,7 @@ function generateDatabaseList(sortedArray, trim) {
     renderValue = h('div.wrap-text', h('ul.db-list', renderValue));
   }
 
-  return h('div.fake-paragraph', [h('div.span-field-name', 'Database References:'), renderValue]);
+  return h('div.fake-paragraph', [h('div.span-field-name', 'Links :'), renderValue]);
 }
 
 //Replace any instances of Replaced 
--- a/src/client/common/tooltips/generateContent.js
+++ b/src/client/common/tooltips/generateContent.js
@@ -41,12 +41,12 @@ const dataSourceHandler = (pair) => {
 };*/
 
 const databaseHandlerTrim = (pair, expansionFunction) => {
-  const expansionLink = h('div.more-link', { onClick: () => expansionFunction(pair[0]) }, 'more »');
+  const expansionLink = h('div.more-link', { onclick: () => expansionFunction(pair[0]) }, 'more »');
   if (pair[1].length < 1) { return h('div.error'); }
   return generateDatabaseList(sortByDatabaseId(pair[1]), true, expansionLink);
 };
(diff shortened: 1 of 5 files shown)


src/client/common/tooltips/generateContent.js
```diff
@@ -3,23 +3,32 @@ const classNames = require('classnames');
 const _ = require('lodash');
 const config = require('./config');
 
+//Define entry limit for trim mode
+const defaultEntryLimit = 3;
```
d2fong: these constants could probably be moved to the config.

hmtinc: I have moved the constants to the config file.

d2fong:
@maxkfranz @jvwong @gaelder @jonahdlin any comments on the visual style/design for the new tooltips?

src/client/common/tooltips/generateContent.js
```diff
@@ -3,23 +3,27 @@ const classNames = require('classnames');
...
-  let shortArray = filterChemicalFormulas(trimValue(pair[1], 3));
   return h('div.fake-paragraph', [
     h('div.field-name', 'Synonyms:'),
-    valueToHtml(shortArray, true)
+    valueToHtml(shortArray, true),
+    expansionLink
   ]);
 };
-const nameHandler = (pair) => {
+const nameHandler = (pair, expansionFunction) => {
```
d2fong: can you describe what expansionFunction is?

src/client/common/tooltips/generateContent.js
```diff
@@ -296,7 +322,7 @@ function noDataWarning(name) {
 function filterChemicalFormulas(list) {
   //Filter out Chemical formulas
   if (list instanceof Array) { return list.filter(name => (!name.trim().match(/^([^J][0-9BCOHNSOPrIFla@+\-\[\]\(\)\\=#$]{6,})$/ig))); }
-  return list;
+  return [list];
```
d2fong: so you are returning an array of arrays here?

jvwong:
I think it looks great. The main things for me are, for entity: what type and what are other names; for process: what type and useful publications.

src/client/common/tooltips/generateContent.js
hmtinc: expansionFunction is the function that is executed when the user clicks on the generated field. It triggers a series of functions which generate the expanded view of the field. 

In the above example, expansionFunction will allow the user to toggle between displaying 3 names and displaying all the names.

src/client/common/tooltips/generateContent.js
hmtinc: @d2fong 
No, the previous line will return if list was an array. Line 235 actually constructs an array of 1 element, so the information can be passed into the generation functions which all require an Array as in a input. 

List is probably a poor choice for a variable name. The function signature should also be Anything -> Array. The reason why this function catches this error is because pc2 will only return a string if there is only one name entry.

gaelder:
I think it looks great too!

jonahdlin:
I like it as well

Issue #263, opened 2017-11-29
Menu bar changes


jonahdlin:
Menu bar changes
- [x]  Change ‘Rearrange Entities’ to ‘Arrange Display’
- [x]  Update display descriptions to be more descriptive
- [x]  Make smart recommendations on the correct layout.
- [x]  Make layouts option available in edit mode only
- [x]  Add a reset button to the menu bar with the default being human created layouts

PR #269, opened 2017-11-29
Disabled extra layouts in view mode

jonahdlin:
Addresses issues in #263, namely

- Change ‘Rearrange Entities’ to ‘Arrange Display’
- Update display descriptions to be more descriptive
- Make layouts option available in edit mode only
- Add a reset button to the menu bar with the default being human created layouts, and if there is none, CoSE-Bilkent

screen shot 2017-11-29 at 2 33 48 pm

### DIFF
--- a/src/client/common/cy/layout/index.js
+++ b/src/client/common/cy/layout/index.js
@@ -1,7 +1,7 @@
 const layouts = [
   {
     displayName: 'Force Directed',
-    description: 'Layout algorithm for undirected compound graphs',
+    description: 'For undirected compound graphs',
     options: {
       name: 'cose-bilkent',
       nodeDimensionsIncludeLabels: true,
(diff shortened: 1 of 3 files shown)


jonahdlin:
The code in the menu bar is a little all over the place, and is too long in general, so I'll be refactoring it in a future PR

jvwong:
This may be nit-picking, but does the icon suggest 'refresh'?

jonahdlin:
I figured that's essentially what the button does, refreshes the graph. What did you have in mind?

d2fong:
Do you mean the icon should look more like?
https://material.io/icons/#ic_refresh

d2fong:
Maybe these are options too.

https://material.io/icons/#ic_autorenew
https://material.io/icons/#ic_loop

jvwong:
I don't have a problem with the icon. With the existing icon, I think the text might be 'Reset network'?

jonahdlin:
I agree, and actually the code for that is already in there. Currently it doesn't show because of the way my tooltips work. When I refactor the tooltips it should change to that.

PR #271, opened 2017-11-30
Search - Removed tabs, changed example, removed database pictures

jonahdlin:
As the title says, got rid of the tabs for different searches and the database pictures next to each search entry. I changed the examples to TP53, Glycolysis, MDM2, but if you want anything else in there, let me know. It looks pretty empty right now, but it is the plan to re-add the tabs at some point, so I figured I'd keep it open.

screen shot 2017-11-30 at 8 58 53 am

Should I change the examples to something else? Should I move the tools button to the left or somewhere else?


jvwong:
is there a way we can test version tabs along the top. in thus case there would be one tab for ' pathways'. that would not be abvious because in fact we are filtering for type pathways.

also yes, adding more search types , like for drugs or interactions is something we will definitely do soon.

jonahdlin:
Sure, I could leave one tab in for pathways, to be expanded in the future. Looks like this now
screen shot 2017-11-30 at 9 20 06 am

d2fong:
@jvwong the current examples are sufficient for you?  e.g glycolysis, tp53..?

jvwong:
I would try to float (at least) three types of input, in order of importance:

1. Phrase / pathway name (e.g. 'cell cycle')
2 Gene(s) of interest (e.g. 'p53 MDM2')
3. Standard biological identifier(s) (e.g. 'P04637')

jonahdlin:
Made those changes. Should be updated now.

PR #272, opened 2017-11-30
Escaping backslashes; blacklisting gene name recognition

jvwong:
1. The backslashes have to be escaped due to 'pathway-commons' npm package escape of backslashes.
2. The 'blacklist' was present in pathways-search but removed, but I put it back because hgnSymbols contains stupid gene symbols alternative names (e.g. 'cell')

### DIFF
--- a/src/server/pathway-commons/search/index.js
+++ b/src/server/pathway-commons/search/index.js
@@ -5,7 +5,7 @@ const getHGNCData = require('./hgnc');
 const sanitize = (s) => {
   // Escape (with '\'), to treat them literally, symbols, such as '*', ':', or space, 
   // which otherwise play special roles in a Lucene query string.
-  return s.replace(/([\!\*\+\-\&\|\(\)\[\]\{\}\^\~\?\:\/\\"\s])/g, '\\$1')
+  return s.replace(/([\!\*\+\-\&\|\(\)\[\]\{\}\^\~\?\:\/\\"\s])/g, '\\\\$1')
 };
 
 const processPhrase = (phrase, collection) => {
@@ -16,15 +16,21 @@ const processPhrase = (phrase, collection) => {
     'refseq'
   ];
 
+  // Do NOT recognize these gene symbols
+  const symbolNameBlackList = [
+    'CELL'
+  ];  
+
   const tokens = phrase.split(/\s+/g);
 
   return tokens
     .map(token => {
       //if symbol is recognized by at least one source
+      const blackListed = symbolNameBlackList.indexOf(token.toUpperCase()) > -1;
       const recognized = sourceList.some(source => utilities.sourceCheck(source, token))
                               || collection.has(token.toUpperCase());
       const sanitized = sanitize(token);
-      return recognized ? ( 'xrefid:' + sanitized ) : ( 'name:' + '*' + sanitized + '*' );
+      return !blackListed && recognized ? ( 'xrefid:' + sanitized ) : ( 'name:' + '*' + sanitized + '*' );
     });
 };
 
--- a/src/server/pathway-commons/search/index.js
+++ b/src/server/pathway-commons/search/index.js
@@ -15,22 +15,16 @@ const processPhrase = (phrase, collection) => {
     'smpdb',
     'refseq'
   ];
-
-  // Do NOT recognize these gene symbols
-  const symbolNameBlackList = [
-    'CELL'
-  ];  
-
   const tokens = phrase.split(/\s+/g);
 
   return tokens
     .map(token => {
       //if symbol is recognized by at least one source
-      const blackListed = symbolNameBlackList.indexOf(token.toUpperCase()) > -1;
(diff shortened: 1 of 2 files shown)


src/server/pathway-commons/search/index.js
```diff
@@ -16,15 +16,21 @@ const processPhrase = (phrase, collection) => {
     'refseq'
   ];
 
+  // Do NOT recognize these gene symbols
+  const symbolNameBlackList = [
```
d2fong: blacklist is checked in here:
https://github.com/PathwayCommons/app-ui/blob/development/src/server/pathway-commons/search/hgnc.js#L4

d2fong: Does it not work without this?

jvwong: nice catch - then for that file, line 15 should be AND rather than OR

jvwong: https://github.com/PathwayCommons/app-ui/blob/development/src/server/pathway-commons/search/hgnc.js#L15

src/server/pathway-commons/search/index.js
```diff
@@ -5,7 +5,7 @@ const getHGNCData = require('./hgnc');
 const sanitize = (s) => {
   // Escape (with '\'), to treat them literally, symbols, such as '*', ':', or space, 
   // which otherwise play special roles in a Lucene query string.
-  return s.replace(/([\!\*\+\-\&\|\(\)\[\]\{\}\^\~\?\:\/\\"\s])/g, '\\$1')
+  return s.replace(/([\!\*\+\-\&\|\(\)\[\]\{\}\^\~\?\:\/\\"\s])/g, '\\\\$1')
```
d2fong: so could you give an example of what the query will look like after this?
the goal is that backslashes should be added to the query?

jvwong: For input 'mu-type opioid' the 'queries' would be the array:

[ '(name:mu\\\\-type\\\\ opioid) OR (name:*mu\\\\-type\\\\ opioid*) OR (name:*mu\\\\-type* AND name:*opioid*)',
  '(name:*mu\\\\-type* OR name:*opioid*)',
  'mu-type opioid' ]

which, in turn, the 'pathway-common' npm package search function code interprets as:
['(name:mu\\-type\\ opioid) OR (name:*mu\\-type\\ opioid*) OR (name:*mu\\-type* AND name:*opioid*)',
...
]

d2fong: So if the pathway-commons npm package does this already, should we just not use sanitize? 

Does it work if you just make sanitize return the original string?

IgorRodchenkov: I disagree with line #8 change - should stay '\\$1', which e.g., replaces:

- ' ' with '\ '
- '\' with '\\'
- ':' with '\:'
- '  ' with '\ \ '
- etc.

which was correct 
(and pathway-commons js client should not do any other tricks but URL-encoding query parameters...)

d2fong: So then it should be changed in the pathway-commons module.

jvwong: I'm suggesting that 'pathway-commons' client does tricks.

Try ['query-string' npm package](https://www.npmjs.com/package/query-string)  function 'stringify': 

```js
var stringify = require('query-string').stringify;
var s = '(name:mu\\-type\\ opioid)';
stringify({q: s}) 

// returns q=(name:mu\-type\ opioid)
```

d2fong: https://github.com/PathwayCommons/cpath2-client/blob/master/src/search.js#L102  Here is where pathway-commons module escapes lucene syntax

d2fong: maybe there is some weird behaviour where qs.stringify and the escapeLucene functions are both being run multiple times on the query...  maybe it might make sense to move away from using the pathway-commons npm package.

d2fong: Max and Igor's proposed new string:

s.replace(/([\/\\!*+-&":|()\[\]{}\^\~\?\s])/g, '\\$1')

d2fong: Manfred lib changes:
- add escape spaces for escapeLucene
- do not escapeLucene if toggle is undefined

https://github.com/PathwayCommons/cpath2-client/blob/master/src/search.js#L102

d2fong: app changes:
- needs to send toggle to the lib to determine if the toggle should be set to true or not

IgorRodchenkov: We've (with @maxkfranz)  got an idea; see PathwayCommons/cpath2-client#12.

d2fong:
I fixed the hgnc filtering bug here 5defbdd3ccd038b0e57c63fcd4eaa78576a5bdba

PR #285, opened 2017-12-04
Database Unit Tests

gaelder:


d2fong:
If someone wants to run the tests do they have to do anything special or is it just ```npm run test```?

gaelder:
I just checked, no npm run test does not work. Should I move the test files to the /test/ instead of organizing them in folders?

d2fong:
What command do you use to run the tests?

gaelder:
I had just been running each file individually with mocha _file_

d2fong:
Do the tests run if you just type ```mocha```?

gaelder:
No

d2fong:
Okay so in package.json, under npm scripts, you can set that command to run ```mocha --recursive```

d2fong:
Can you paste the output of just running ```mocha``` it is strange that you need the recursive flag for all the tests to run.

gaelder:
This is the output.
`  Example
    ✓ does something
    ✓ does something else


  2 passing (11ms)`

I have my tests in a subdirectory of /test because I figured I should be mirroring the directory structure of /src, hence the need for the flag

d2fong:
https://stackoverflow.com/questions/10753288/how-to-specify-test-directory-for-mocha

Can you try something similar to:
```mocha "test/**/*.js"```

gaelder:
```mocha "test/**/*.js``` has the same output as ```mocha --recursive```

d2fong:
Okay so I guess they both do something similar.

PR #292, opened 2017-12-06
Edit warning changes

jonahdlin:
Most importantly, this fixes an annoying bug with the edit warning that appears in the edit mode. I also made it into its own (quite limited) module, changed the visual style to be more consistent and added a dismiss button to it like in factoid. I also attempted to fix a similar bug with the sidebar, but the bug is hard to produce in the first place, so it's hard to tell if it's fixed.

screen shot 2017-12-06 at 2 28 21 pm


d2fong:
Nice. Good job.

PR #305, opened 2017-12-08
Live updates

jonahdlin:
This PR adds the live update collaboration functionality that we attempted earlier in the term. Now two people in edit mode on the same pathway can see changes occurring in real time, like in google docs. This doesn't change any of the layout saving functionality.

### DIFF
--- a/src/server/routes/sockets.js
+++ b/src/server/routes/sockets.js
@@ -5,6 +5,12 @@ const qs = require('querystring');
 const io = require('./../io').get();
 
 io.on('connection', function (socket) {
+  let referer = socket.handshake.headers.referer;
+  let params = referer.match(/(edit|view)\?(.*)/);
+  if (params){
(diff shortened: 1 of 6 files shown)


src/server/routes/sockets.js
```diff
@@ -5,6 +5,12 @@ const qs = require('querystring');
 const io = require('./../io').get();
 
 io.on('connection', function (socket) {
+  let referer = socket.handshake.headers.referer;
```
d2fong: Can you elaborate on what this chunk of code is doing?

jonahdlin: @gaelder Can handle this one. He wrote the backend parts of this PR.

gaelder: @d2fong This grabs the url that the client is currently viewing. It is used to determine which pathway is being viewed.

PR #307, opened 2017-12-08
Added a generate share link button

hmtinc:
It was suggested that there should be a way of sharing the specific view the user is focusing including zoom, pan, and temp node positions. 

### Client Code
To address this, we have implemented snapshot feature on the front end called `Get Shareable Link' which will take a snapshot of the current view and return a URL to the user to share with other users. 

##Server Side Code
**Geoff** wrote a few api functions that will store the snapshot and return a processable id back to the front end. This API does not modify any of the main data entries in the database, as a result we have decided to make the snapshot feature publicly accessible 

## Example
example


src/client/features/view/components/menu/share.js
```diff
@@ -0,0 +1,41 @@
...
+ * @desc Gets a sharable link and saves the link to the clipboard. 
+ */
+function getShareLink(cy, uri) {
+  //Get graph information
+  const nodes = cy.nodes();
+  const pan = cy.pan();
+  const zoom = cy.zoom();
+
+  //Get base url
+  let baseUrl = window.location.href;
```
d2fong: You should be using react router for this stuff

d2fong:
Keep this pr open but submit a new pr to the demo branch

PR #311, opened 2017-12-08
Version fix

gaelder:
duplicate of #291 but into the demo branch.


d2fong:
What should the conflict resolution be @hmtinc @gaelder

gaelder:
@d2fong resolved.

PR #315, opened 2017-12-08
Added Base Name Option and Global Config

### LINKED ISSUE #276: Using HashRouter instead of BrowserRouter?
@d2fong  @maxkfranz @IgorRodchenkov 

Igor wants to host the application in a virtual directory, so he is able to have multiple PC apps/services on one domain. i.e `http://pathwaycommons.org/apps/search' for search . 

This causes issues with BrowserRouter though, as the full path is used by the router to determine the correct page. i.e `/apps/search/view`  should be `/view`.  This issue can easily be mitigated by defining a **basename** `{ basename: '/apps/search' }`, but this would require a future user to change the **basename** if they want to deploy the site anywhere else (locally, another folder, or another server). 

Using HashRouter fixes this issue as it separates the actual server path from the react route with a hash (#), thus the app can be moved anywhere without having to define a base directory name.  

I know hashrouter typically shouldn't be used for a dynamic site and that it complicates the URL,  but it seems alot more ideal than having to go into the code and manually define the base directory whenever you want to host the application elsewhere. 

Given this, should we change BrowserRouter to HashRouter, or is there a more ideal solution?

hmtinc:
closes #276 - Added a global config file with a baseName field that is set based on the BASE_NAME environment variable.

This should allow the application to be hosted from any directory as long as the BASE_NAME environment variable is defined to the matching directory path.


d2fong:
Merged it to the wrong branch

d2fong:
create this pr again for the demo branch

d2fong:
@hmtinc

PR #324, opened 2017-12-11
Small browser compatibility fixes

jonahdlin:
Fixed an issue with padding in Firefox, and added a nasty patch job to make the menu bar behave a little better in Safari. It would seem Safari doesn't bother calculating the width of inner components of absolutely positioned divs, meaning the toolbar ends up looking like this in Safari:

screen shot 2017-12-11 at 9 28 08 am

The fix I made here doesn't fix that, but makes the animations work smoothly, whereas they used to be jerky.

Any ideas about how to fix the wide toolbar in Safari?


d2fong:
I got nothing.

PR #329, opened 2017-12-11
DEMO: Share

jonahdlin:
Revamped the snapshot button, and modelled it after the github clone link. It now appears in the sidebar along with the dropdown and search bar.

screen shot 2017-12-11 at 3 13 41 pm
screen shot 2017-12-11 at 3 13 32 pm

I also moved the CSS files in the common folders out of their folders, since the folder names were identical to the component names.

This is based off the share link code written by Harsh and pulled into demo, so when/if this is pulled into development there will likely be a lot of merge conflicts to deal with.


maxkfranz:
If the text box is going to be that small, is it worth having?  Maybe we could just copy the url to the clipboard when clicking the share (three node) icon.

Aside : The icon is a bit confusing, since it looks like a graph and much of the ui is a graph.  That's one icon that google did a poor job on imo.  Maybe one of these?

Link : https://material.io/icons/#ic_link <-- I use this on on factoid

Export / external link : https://material.io/icons/#ic_exit_to_app

Variants of above : https://material.io/icons/#ic_launch , https://material.io/icons/#ic_open_in_browser

jonahdlin:
I thought about doing that but I figured it could be intrusive to overwrite the clipboard without it being obvious to the user that it was going to happen.

I changed the icon to the factoid one (link).

gbader:
I agree with Max that the text box is too much.  If we want to copy to clipboard, we should find a way that doesn't take up any toolbar real estate - it's jut not worth it for that convenience feature.  For instance, you could pop up another tooltip that has the sharing UI, including a button to copy to clipboard.

jonahdlin:
I've pushed the following change. By sharing UI, do you mean the full thing with the text box, or just with a button?
screen shot 2017-12-15 at 1 24 00 pm

gbader:
looks great! thanks.

PR #330, opened 2017-12-12
Pulled controller code out of the rest functions

gaelder:


src/server/routes/controller.js
```diff
@@ -100,12 +104,50 @@ function renderPNG(cyJson) {
   return renderImage(cyJson);
 }
 
+function purgeSnapshot() {
+  const tenDays = 10 * 60 * 60 * 24;
```
hmtinc: Number of days should be a parameter taken in by purgeSnapshot()

hmtinc: Number of days should be a parameter taken in by purgeSnapshot()

d2fong:
So how does purging work?  Someone has to manually call it?

gaelder:
It has a rest endpoint. I have a cronjob set to call it daily on my local machine, though this job needs to be created anywhere the server is run.

d2fong:
Okay, could you also explain the reasoning for storing snapshots?  I think I missed that discussion.

d2fong:
Okay makes sense.  Thanks.

hmtinc:
@d2fong  Its somewhat equivalent to a temporary "edit" feature. It allows regular "view" users to make modifications and share their modifications with others without effecting that actual official layout of the graph.

Issue #321, opened 2017-12-11
Add expand/collapse option in the tooltip


d2fong:

hmtinc:
@d2fong Given that the tooltip is meant for information, it feels that adding a expand/collapse button would over-complicate the tooltip. How about adding a context menu in addition to the tooltip?

![screen shot 2017-12-11 at 9 37 53 am](https://user-images.githubusercontent.com/1313810/33836551-b1d62ea2-de57-11e7-9ef8-95f1a43b5a37.png)

jvwong:
I vote yes on the context-menu. I think there are additional functionalities that would be put here in the future so good to have it ready. For example, edit this pathway via [Newt](https://github.com/iVis-at-Bilkent/newt/issues/187), search other pathways with this gene (Igor has this info ready)

gbader:
The problem with right click is that it's not easily visible and as a result I don't think many people will find and use it. Right click is good for power user apps that have a lot of functionality.  Our apps are much simpler and if we get to the high number of features that need a right click menu, then I would question whether we need all of those features for an app.  It's also not good to make things more complex for potential future features - best to only do that once we need those features, otherwise it's over-engineering (bad).  GeneMANIA handles this by adding a couple of buttons on the tooltip (e.g. for adding a node to a query), but even then, I doubt people are using this functionality. But even then I doubt people use this because it's easier to just issue a new query.

jvwong:
I would be fine with disabling this in the UI (but keeping it in the code).

PR #331, opened 2017-12-13
Context Menu

hmtinc:
#321 

Jeff suggested adding a context menu to allow for additional options beyond expand/collapse and show tooltip.

I have mocked up a rough tooltip using the cytoscape-cxtmenu extension used in factoid to just demonstrate the features potential use case. 

@jvwong @d2fong  @maxkfranz , any suggestions on potential design changes. 

screen shot 2017-12-13 at 9 38 12 am
 
Note : Merging this into the 'fun' branch just to seek feedback on the idea of having a context menu.

### DIFF
--- a/src/client/features/view/components/graph/contextMenu.js
+++ b/src/client/features/view/components/graph/contextMenu.js
@@ -1,0 +1,52 @@
+const _ = require('lodash');
+
+const defaults = {
+  menuRadius: 80,
+  fillColor: 'rgba(0, 0, 0, 0.75)',
+  activeFillColor: 'rgba(1, 105, 217, 0.75)',
+  activePadding: 10,
+  indicatorSize: 12,
+  separatorWidth: 3,
+  minSpotlightRadius: 12,
+  maxSpotlightRadius: 12,
+  openMenuEvents: 'cxttapstart taphold',
+  itemColor: 'white',
+  itemTextShadowColor: 'transparent',
+  zIndex: 9999,
+  atMouse: false
+};
+
+const expandCollapse = generateCommand('settings_overscan', 'Expand/Collapse', ele => {
+  ele.emit('hideTooltips');
+  ele.emit('expandCollapse');
+});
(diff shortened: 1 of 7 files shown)


src/client/features/view/components/graph/contextMenu.js
```diff
@@ -0,0 +1,52 @@
...
+const getInfo = generateCommand('info_outline', 'More Info', ele =>  ele.emit('showTooltip'));
+
+//Create a cxt menu command object
+function generateCommand(icon, title, selectFunction) {
+  return {
+    content: '<i class="material-icons cxt-icon">' + icon + '</i> <div class="cxt-info">' + title + '</div>',
+    select: selectFunction
+  };
+}
+
```
d2fong: I think it makes sense to move this file to common/cy/events/  Is there a particular reason you think it should go in components/graph?

d2fong:
Current discussion seems divided on whether or not to use context menus.  In terms of design maybe have the context select colour match the deep sea green highlight colour to match?

jvwong:
Can we simply disable in UI, then merge?

src/client/features/view/components/graph/contextMenu.js
hmtinc: The actual file could be moved. As for the binding of events, I found that binding before elements were added to the graph caused errors with the extension which is why I bound he events in /graph/index.js

hmtinc:
@jvwong do you want it disabled and merged with the development branch or this 'fun' branch?

jvwong:
Lets leave it in fun. Wasn't very popular idea.

d2fong:
The code will still live on in the fun branch.  If discussions later down the line suggest we should use the context menu, then it can be merged back to development.

d2fong:
@hmtinc @gaelder @jonahdlin feel free to merge anything you want into this fun branch.  You can do whatever you want in here.

PR #336, opened 2017-12-16
reusable base view component

d2fong:
Note:  Early Jan. merge to dev.  (won't be merged for initial pc dev/mailing list send out)

Refactor most of the view code to be a common view component.  
- Layout configuration and component configuration(view/edit/paint/networks in the future) inject their specific buttons and menus into the base component.
- easier to develop features for view/paint/edit specifically as there is a proper place to put these features while still reusing the main view template
- paint/edit/view all use the base view component

Casualties of this large refactor / TODO:
- ~~~multi-edit is broken now, oops...~~~ EDIT: not anymore, turns out I should be accessing updates via the ```nodeID``` key instead of the ```nodeId```.  Should probably be fixed in the future.
- ~~~I got rid of all the other layouts except cose-bilkent and human created, I just don't think they will be useful now that complex nodes are expanded by default. another benefit to this is a decreased bundle size~~~ Will add them back now that I think I know how to specify injecting non-button components
- still have to think about how one would inject more complicated components into the tool bar, such as the node search bar and the layout dropdown
- create a shell template for the loading state
- make the sidebar width depend on the current menu width
- document the interface / how to use this base component
- potentially hook up an event bus to make things even more simpler

Solution is not perfect, but this seemed like the easiest way to have code reuse.  Took a while but now the paint app looks like the view app and uses it's own paint button that is injected into the template:

screen shot 2017-12-15 at 8 00 49 pm


d2fong:
might also want to think about creating a light wrapper over the menu/button registration system.

e.g base view component keeps internal representation of button->menu map
```js
// internal representation
buttons = [...]
menus = [...]
```

```js
// method exposed to view/paint/networks/edit
// requires button spec to have X properties, menu to be a react hyperscript component function
registerMenu(buttonSpec, menu) 
```

d2fong:
It might make sense to do away with this buttonSpec business and just register a buttonComponent too.  Being able to register components in the main menu bar would also generalize to other components like dropdowns and input bars, but I don't know how the css should work with these things.

d2fong:
freeze the view config, make clients create an extended copy

d2fong:
Might also make sense to just let edit/view/paint/networks define their own toolbar items and pass them to the toolbar as props

PR #338, opened 2017-12-18
Konami Code Easter Egg

hmtinc:
Added a fun little konami code easter egg. Entering the konami code `up up down down left right left right b a` brings up a special Pathway Commons Gene which displays the credits for the entire pathways commons project. 

example

Note : For the fun branch only


d2fong:
Cool easter egg.  Maybe you could add a node for each team member with a picture over the node bg and connect them to the pc node.

PR #384, opened 2018-01-11
All download options now appear at once.

logmosier:
Issue #373. All download options now appear at once. Also removed hidden option from downloadTypes as it is no longer used.


d2fong:
You should be able to move issues in your project board now.

d2fong:
You should also have the permissions to close these issues yourself.

PR #391, opened 2018-01-18
Hide node search bar 'close' icon when there is no search bar input

wxli0:


jvwong:
I'm still seeing the close icon.

image

wxli0:
I'm trying to figure out what's going wrong.

d2fong:
ok it should work now #394

PR #407, opened 2018-01-25
initial commit

wxli0:
@d2fong Dylan, I added two simple tests at test/gene-query-test.js. Could you please check if i'm on the right track?


d2fong:
Yes this is a good start.

d2fong:
Do these tests pass?

wxli0:
Since I haven't implemented the validate function (step 4 on my to-do list), none of them passes.

src/server/routes/rest.js
```diff
@@ -36,6 +37,12 @@ router.get('/get-graph-and-layout', function (req, res) {
   });
 });
 
+router.get('/test/', (req, res) => {
```
d2fong: This route should be renamed.  Instead of test is should be 'gene-query'

src/server/routes/rest.js
```diff
@@ -36,6 +37,12 @@ router.get('/get-graph-and-layout', function (req, res) {
...
+  const genes = req.query.genes;
+
+  res.json(geneQueryService(genes));
```
d2fong: you should return an empty response since geneQueryService has not been implemented.  This will throw an error if you go to localhost:3000/api/test

d2fong:
Ok that makes sense.  Just make the changes I specified before you move on.

d2fong:
Also in the future you need to make commit messages more relevant.

e.g instead of initial commit it should be 'made tests for gene query service'

src/server/routes/rest.js
wxli0: What's an empty response?

d2fong: return an empty object as the response.


res.json({})

PR #408, opened 2018-01-26
398

wxli0:


src/server/enrichment-map/index.js
```diff
@@ -1 +0,0 @@
-module.exports = {};
```
d2fong: You should revert this change

d2fong: Should be gene-query/index.js

src/server/routes/rest.js
```diff
@@ -36,6 +37,11 @@ router.get('/get-graph-and-layout', function (req, res) {
   });
 });
 
+router.get('/gene-query/', (req, res) => {
+  const genes = req.query.genes;
+
+  res.json(geneQueryService({}));
```
d2fong: This should be res.json({})

src/server/tokenize.js
```diff
@@ -0,0 +1,5 @@
+const tokenize = function(input) {
```
d2fong: This should be gene-query/tokenize

PR #409, opened 2018-01-26
399

wxli0:


src/server/mapping/mapping.js
```diff
@@ -0,0 +1,22 @@
...
+const fs = require("fs");
+const path = require("path");
+const jsonData = fs.readFileSync(path.resolve(__dirname, "HGNC_dataset.json"));
+const jsonContent = JSON.parse(jsonData);
+
+
+// map HGNC symbol to HGNC ID
+// After validation step, we can assume all symbols are valid
+// so that we can find a mapping for sure
+const mapping = function(symbol) {
```
d2fong: make names more descriptive and less general 

```symbol``` changed to ```hgncSymbol```

File: src/server/mapping/mapping.js [hunk 2]
```diff
@@ -0,0 +1,22 @@
...
+  let id = 0;
```
d2fong: So if the symbol isnt found you return 0?

I think it would be better to return null

wxli0: Since we call validate before mapping, all symbols will be found for sure.

Yes. I agree it'd be better to return null.

d2fong:
@wxli0 can you rebase this branch ontop of development?

PR #412, opened 2018-01-30
Search info box

logmosier:
The search info box.


src/client/services/server-api/index.js
```diff
@@ -28,6 +28,20 @@ const ServerAPI = {
...
+  return fetch(`/pc-client/querySearch?${qs.stringify(query)}`, defaultFetchOpts).then(res => res.json()).then(res=> {
+    return _.compact(res.map(hit=>{
+        if( hit.uri.startsWith('http://identifiers.org/uniprot/') && _.endsWith(hit.organism[0],'9606')){
+         return _.last(hit.uri.split('/'));
+        }
+      }));
+    });
+  },
+
+  landingBox(query){
```
d2fong: Does this function get protein information based on some query args?

logmosier: It gets the protein information from UniProt based on the id that is passed in.

d2fong: Okay in that case it might make sense to rename this to

```js
getProteinInformation(uniprotId)
```
1.  ```landingBox``` implies that this function will only be used for the landing box but really anything could potentially want this information.  

2. ```query``` is a pretty general argument name.  In js it is more important to be a bit more specific with names.

src/client/features/search/index.js
```diff
@@ -200,6 +267,7 @@ class Search extends React.Component {
       h(Loader, { loaded: !state.loading, options: { left: '50%', color: '#16A085' } }, [
         h('div.search-list-container', [
           h('div.search-result-info', [searchResultInfo]),
+          h('div',{style:{width:'100%'}},[landing]),
```
d2fong: I think this should be moved to the search.css file

src/client/features/search/index.js
```diff
@@ -148,6 +182,39 @@ class Search extends React.Component {
     ]) :
       h('div.search-hit-counter', `${state.searchResults.length} result${state.searchResults.length === 1 ? '' : 's'}`);
 
+    const landing = (state.landingLoading && state.searchResults.length>0) ?
+      h('div.search-landing.innner',[h(Loader, { loaded:!state.landingLoading , options: { color: '#16A085',position:'relative', top: '15px' }})]):
+      state.landing.map(box=>{
+        const synonyms=_.hasIn(box,'protein.alternativeName') ? 
+          h('i.search-landing-small',box.protein.alternativeName.map(obj => {return obj.fullName.value;}).join(', ')):'';
+
+        const showFunction = state.landingShowMore ?  '.search-landing-showContent' : '.search-landing-hideContent';
```
logmosier: can use classNames package to simplify conditional css class assignment

src/client/services/server-api/index.js
```diff
@@ -28,6 +28,20 @@ const ServerAPI = {
     return fetch(`/pc-client/querySearch?${qs.stringify(query)}`, defaultFetchOpts).then(res => res.json());
   },
 
+  findUniprotId(query){
+  return fetch(`/pc-client/querySearch?${qs.stringify(query)}`, defaultFetchOpts).then(res => res.json()).then(res=> {
+    return _.compact(res.map(hit=>{
+        if( hit.uri.startsWith('http://identifiers.org/uniprot/') && _.endsWith(hit.organism[0],'9606')){
```
logmosier: instead of hardcoding 9066 (human), pass this in as a query param

src/client/services/server-api/index.js
```diff
@@ -28,6 +28,20 @@ const ServerAPI = {
...
+         return _.last(hit.uri.split('/'));
```
logmosier: maybe add a comment saying that you are parsing the uniprot uri

src/client/features/search/index.js
```diff
@@ -42,19 +45,51 @@ class Search extends React.Component {
...
+    const state = this.state;
+    const query = {
+      q: state.query.q,
+        gt: -1,
+        lt: 250,
+        type: 'ProteinReference',
+        datasource: state.query.datasource,
+        species: '9606'
+    };
+    this.setState({
```
d2fong: setState is asynchronous. to gaurantee that the code after this runs in sequence you should do something like:

```js
this.setState({
  landingLoading: true
}, () => {
  ServerAPI.findUniprotId(...)...
});

```

d2fong: you can optionally supply a callback function that runs after setState has completed

src/client/features/search/index.js
```diff
@@ -148,6 +183,39 @@ class Search extends React.Component {
     ]) :
       h('div.search-hit-counter', `${state.searchResults.length} result${state.searchResults.length === 1 ? '' : 's'}`);
 
+    const landing = (state.landingLoading && state.searchResults.length>0) ?
+      h('div.search-landing.innner',[h(Loader, { loaded:!state.landingLoading , options: { color: '#16A085',position:'relative', top: '15px' }})]):
+      state.landing.map(box=>{
+        const synonyms=_.hasIn(box,'protein.alternativeName') ? 
+          h('i.search-landing-small',box.protein.alternativeName.map(obj => {return obj.fullName.value;}).join(', ')):'';
```
d2fong: you can omit the curly brackets and return statement in the map function:

```js
alternativeName.map(obj => obj.fullName.value)
```

Issue #369, opened 2018-01-04
Search result -- nothing displayed. TypeError: Cannot read property 'name' of undefined


ariutta:
Hello,

I tried searching for tp53, but there was an error that meant the display was blank.

URL: http://beta.pathwaycommons.org/search?gt=0&lt=250&q=tp53&type=Pathway
Browser: Chrome Version 63.0.3239.84
OS: macOS Sierra

Error:
> deps.js:36 TypeError: Cannot read property 'name' of undefined
    at bundle.js:1
    at Array.map (<anonymous>)
    at t.value (bundle.js:1)
    at s (deps.js:36)
    at beginWork (deps.js:36)
    at i (deps.js:36)
    at a (deps.js:36)
    at u (deps.js:36)
    at x (deps.js:36)
    at w (deps.js:36)
s @ deps.js:36
bundle.js:1 Uncaught (in promise) TypeError: Cannot read property 'name' of undefined
    at bundle.js:1
    at Array.map (<anonymous>)
    at t.value (bundle.js:1)
    at s (deps.js:36)
    at beginWork (deps.js:36)
    at i (deps.js:36)
    at a (deps.js:36)
    at u (deps.js:36)
    at x (deps.js:36)
    at w (deps.js:36)

jvwong:
I traced this to undefined variable `dsInfo` in [app-ui/src/client/features/search/index.js](https://github.com/PathwayCommons/app-ui/blob/3d1e1bdac4dfd0d7ce60067f85306ff4a4ae7ddd/src/client/features/search/index.js#L106-L120) 
```js
  const searchResults = state.searchResults.map(result => {
      const dsInfo = _.find(state.dataSources, ds => {
        return ds.uri === result.dataSource[0];
      });

      return h('div.search-item', [
        h('div.search-item-content', [
          h(Link, { to: { pathname: '/view', search: queryString.stringify({ uri: result.uri }) }, target: '_blank' }, [
            h('h3.search-item-content-title', result.name || 'N/A'),
          ]),
          h('p.search-item-content-datasource', ` ${dsInfo.name}`),
          h('p.search-item-content-participants', `${result.numParticipants} Participants`)
        ])
      ]);
    });

```

[...]

PR #413, opened 2018-01-30
Issue369

logmosier:
Fix for Issue #369. Was not able to replicate error but used setTimeout to generate the conditions of the error and fixed that.


src/client/features/search/index.js
```diff
@@ -104,21 +104,26 @@ class Search extends React.Component {
...
-      });
-      return h('div.search-item', [
-        h('div.search-item-icon',[
-          h('img', {src: dsInfo.iconUrl})
-        ]),
-        h('div.search-item-content', [
-          h(Link, { to: { pathname: '/view', search: queryString.stringify({ uri: result.uri }) }, target: '_blank' }, [
-            h('h3.search-item-content-title', result.name || 'N/A'),
+      if(_.isEmpty(state.dataSources)){
+        return h('div');
```
logmosier: might make sense to keep returning relevant info such as title, etc, but show default datasource data if it couldnt be loaded (blank image, omit datasource name in the search hit)

PR #415, opened 2018-02-01
Gene query tests

wxli0:


src/server/mapping/mapping.js
```diff
@@ -1,16 +1,18 @@
 
 const fs = require("fs");
 const path = require("path");
-const jsonData = fs.readFileSync(path.resolve(__dirname, "Validation_test_data_set.json"));
+const jsonData = fs.readFileSync(path.resolve(__dirname, "HGNC_dataset.json"));
```
d2fong: Move this file to gene query folder

src/server/gene-query/mapping.js
```diff
@@ -0,0 +1,22 @@
...
+const jsonData = fs.readFileSync(path.resolve(__dirname, "HGNC.json"));
+const jsonContent = JSON.parse(jsonData);
+
+
+// map HGNC symbol to HGNC ID
+// After validation step, we can assume all symbols are valid
+// so that we can find a mapping for sure
+const mapping = function(hgncSymbol) {
+  let id = null;
+  jsonContent.forEach(element => {
```
d2fong: It probably makes sense to transform jsonContent into a map https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map

so you can just lookup symbols.

src/server/gene-query/validate.js
```diff
@@ -0,0 +1,57 @@
...
+  listOfToken.forEach(element => {
+    if(!mapping(element.symbol)) {
+      element.legal = false;
+    } else {
+      element.id = mapping(element);
+    }
+  });
+
+  // check duplicates
+  listOfToken.forEach(element => {
```
d2fong: delegate collection processing to lodash  https://lodash.com/docs/4.17.4#uniqBy

src/server/gene-query/mapping.js
wxli0: Since we have many objects, do you mean an array of map?

d2fong: right now you are iterating through each jsonContent element and checking if element['approved_symbol'] matches the given hgncSymbol

what you could do is let jsonContent be a map so you have constant time lookup for a given symbol

e.g

```js
const jsonData = JSON.parse(s.readFileSync(path.resolve(__dirname, "HGNC.json")));
const jsonContent = new Map();
jsonContent.forEach(el => jsonContent.set(el['Approved_Symbol'], element));

// then in mapping
const validSymbol = jsonContent.has(hgncSymbol)

If (validSymbol) {
 ...
} else {
...
}

PR #420, opened 2018-02-02
Endpoint

wxli0:
Hi Dylan, I think I'm getting closer but still having some issues with the endpoint. The error message didnt make sense to me. It'd be helpful if we can go through the code together sometime and figure out what's going wrong. Thanks.


d2fong:
reopen this pr in development

maxkfranz:
@d2fong We probably should `git merge --squash student-branch` when we merge these into dev.  Otherwise we'll get all the intermediary commit messages in the history, which aren't that useful to keep.

PR #441, opened 2018-02-20
Interaction view

logmosier:
Basic interaction view that fetches the neighborhood on the client side then cuts the neighborhood to 1 degree (makes the network somewhat more reasonable in size), and colours the edges based on the interaction type.

### DIFF
--- a/src/client/common/components/base-network-view/base-network-view.js
+++ b/src/client/common/components/base-network-view/base-network-view.js
@@ -56,7 +56,11 @@ class BaseNetworkView extends React.Component {
     cy.mount(container);
     cy.remove('*');
     cy.add(state.networkJSON);
-
+    if(state.interactionView){
+      const mainNode=cy.nodes(node=> node.data().id===state.id);
+      const nodesToKeep=mainNode.merge(mainNode.connectedEdges().connectedNodes());
+      cy.remove(cy.nodes().difference(nodesToKeep));
+    }
     const layout = cy.layout(initialLayoutOpts);
     layout.on('layoutstop', () => {
       cy.emit('network-loaded');
--- a/src/client/common/components/base-network-view/base-network-view.js
+++ b/src/client/common/components/base-network-view/base-network-view.js
@@ -56,11 +56,7 @@ class BaseNetworkView extends React.Component {
     cy.mount(container);
     cy.remove('*');
     cy.add(state.networkJSON);
-    if(state.interactionView){
-      const mainNode=cy.nodes(node=> node.data().id===state.id);
-      const nodesToKeep=mainNode.merge(mainNode.connectedEdges().connectedNodes());
-      cy.remove(cy.nodes().difference(nodesToKeep));
-    }
+    cy.emit('trim');
     const layout = cy.layout(initialLayoutOpts);
     layout.on('layoutstop', () => {
       cy.emit('network-loaded');
--- a/src/client/common/cy/stylesheet.js
+++ b/src/client/common/cy/stylesheet.js
@@ -40,6 +40,35 @@ const stylesheet = sbgnStyleSheet(cytoscape)
 .selector('edge')
 .css({
   'opacity': 0.3
+})
+.selector('edge[class="Phosphorylation"]')
+.css({
+  'line-color': 'red'
+})
+.selector('edge[class="Binding"]')
+.css({
+  'line-color': 'green'
+})
+.selector('edge[class="Expression"]')
+.css({
+  'line-color': 'blue'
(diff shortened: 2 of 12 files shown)


src/client/features/interactions/index.js
```diff
@@ -0,0 +1,135 @@
...
+const { ServerAPI } = require('../../services/');
+
+const { BaseNetworkView } = require('../../common/components');
+const { getLayoutConfig } = require('../../common/cy/layout');
+
+class Interactions extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      cy: make_cytoscape({ headless: true }),
```
d2fong: you should:
* create a new stylesheet for interactions specifically and inject it into this make cytoscape function
* modify the make cytoscape function to look for a ```stylesheet``` arg and send that arg to the cytoscape constructor if found. Default to the sbgn stylesheet if not found

src/client/common/cy/stylesheet.js
```diff
@@ -40,6 +40,35 @@ const stylesheet = sbgnStyleSheet(cytoscape)
...
+  'text-outline-width': 2,
+  'text-wrap': 'wrap',
+  'text-max-width': 175,
+  'width': 45,
+  'height': 45,
+  'label': node => {
+    const label = node.data('label')
+      .split('(').join('').split(')').join('')
+      .split(':').join(' ');
+    return label;
```
maxkfranz: Would you please comment what's going on here.  Could we memoize the result?  All that splitting and joining is expensive.

src/client/features/interactions/index.js
```diff
@@ -0,0 +1,135 @@
...
+  }
+
+  pathwayLinks(line){
+    return line[6].split(';').map( link => {
+      const splitLink=link.split('/').reverse();
+      return [splitLink[1]==='reactome'? 'reactome': 'Pathway Commons',splitLink[0]];
+    });
+  }
+
+  addInteraction(splitLine,network,nodeMap){
```
maxkfranz: `parse()` should be handling all parsing.  Functions like `addInteraction()` should receive the already parsed data.

E.g. `splitLine` should be replaced by something like `nodeAId, nodeBId`

src/client/common/cy/stylesheet.js
maxkfranz: Couldn't you just use a regex to get the parts you want?

d2fong: Ah thats some legacy stuff related to complex label processing.  That can probably be deleted.

src/client/features/interactions/index.js
```diff
@@ -0,0 +1,129 @@
...
+      id: nodes[0]+edge+nodes[1] ,
+      label:nodes[0]+' '+edge+' '+nodes[1] ,
+      source:nodes[0],
+      target: nodes[1],
+      class: this.edgeType(edge),
+      parsedMetadata:[['Database IDs',sources]]
+    }});
+  }
+
+  parse(data,query){
```
d2fong: what does this parse function do?

logmosier: Takes in the fetched data, and the gene that was searched for and returns an object of the network cut to 1 degree that BaseNetworkView can display.

src/client/common/components/base-network-view/base-network-view.js
```diff
@@ -56,7 +56,11 @@ class BaseNetworkView extends React.Component {
     cy.mount(container);
     cy.remove('*');
     cy.add(state.networkJSON);
-
+    if(state.interactionView){
```
d2fong: Instead of adding an explicit if check can you try emitting another event such as 'json-loaded' and then running the network pruning code on the event?  It might not work but if it does it would be preferable to if checks.

here is an example of where feature specific events are handled:

https://github.com/PathwayCommons/app-ui/blob/development/src/client/features/edit/index.js#L57

src/client/common/components/base-network-view/base-network-view.js
```diff
@@ -56,7 +56,11 @@ class BaseNetworkView extends React.Component {
...
+      const mainNode=cy.nodes(node=> node.data().id===state.id);
+      const nodesToKeep=mainNode.merge(mainNode.connectedEdges().connectedNodes());
+      cy.remove(cy.nodes().difference(nodesToKeep));
+    }
     const layout = cy.layout(initialLayoutOpts);
     layout.on('layoutstop', () => {
       cy.emit('network-loaded');
```
d2fong: here is an example of emitting a custom event

Issue #443, opened 2018-02-21
Infobox: Links


jvwong:
Suggestion is to [alter infobox links](https://github.com/PathwayCommons/app-ui/blob/4fcde7169ec4618383146a0d764307853370a26f/src/client/features/search/index.js#L211-L213) to include only: 

- UniProt
- [HGNC](https://www.genenames.org/)
- Entrez [Gene](https://www.ncbi.nlm.nih.gov/gene)

---

I noticed that UniProt object dbReferences has most of these IDs...

Issue #442, opened 2018-02-21
Include 'More>>' alternative names (synonyms) for protein  


jvwong:
Currently, the info box [concatenates protein.alternativeNames from UniProt then limits either the  total characters/names](https://github.com/PathwayCommons/app-ui/blob/4fcde7169ec4618383146a0d764307853370a26f/src/client/features/search/index.js#L198). 

Suggestion is to simply hide other alternativenames using 'more' option.

---

The alternativeNames array is likely ordered alphabetically, causing some common synonyms to be removed. (e.g. [CDKN1A](http://localhost:3000/search?gt=0&lt=250&q=CDKN1A&type=Pathway) drops 'p21')

![screenshot at feb 21 15-43-27](https://user-images.githubusercontent.com/4706307/36504519-c5cc48b0-171e-11e8-8cd1-20794a00c017.png)

PR #445, opened 2018-02-22
Search info box

logmosier:
Made changes in issue #442 and #443.

### DIFF
--- a/src/client/features/search/index.js
+++ b/src/client/features/search/index.js
@@ -90,7 +90,7 @@ class Search extends React.Component {
         }
         else{
           this.setState({
-            landingLoading: [false,false],
+            landingLoading: false,
             landing:[]
           });
         }
--- a/src/client/features/search/index.js
+++ b/src/client/features/search/index.js
@@ -30,7 +30,7 @@ class Search extends React.Component {
       searchResults: [],
       loading: false,
       showFilters: false,
-      landingShowMore: false,
+      landingShowMore: [false,false],
       dataSources: []
     };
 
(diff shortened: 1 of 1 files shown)


src/client/features/search/index.js
```diff
@@ -30,7 +30,7 @@ class Search extends React.Component {
       searchResults: [],
       loading: false,
       showFilters: false,
-      landingShowMore: false,
+      landingShowMore: [false,false],
```
d2fong: what is the purpose for using two values here?

src/client/features/search/index.js
```diff
@@ -192,26 +192,42 @@ class Search extends React.Component {
     const landing = (state.landingLoading ) ?
       h('div.search-landing.innner',[h(Loader, { loaded:!state.landingLoading , options: { color: '#16A085',position:'relative', top: '15px' }})]):
       state.landing.map(box=>{
+
         let synonyms=null;
         if(_.hasIn(box,'protein.alternativeName')){ 
-          const synonymsText= box.protein.alternativeName.map(obj => obj.fullName.value).join(', ');
-          synonyms=h('i.search-landing-small',synonymsText.slice(0,synonymsText.indexOf(',',70)));
+          const synonymsLength=115;
```
d2fong: I think all of this logic about hiding/showing more information can be handled by css.  We can go over this in person.

src/client/features/search/index.js
logmosier: One for each of the show more options.

PR #447, opened 2018-02-22
Validator info

wxli0:


src/client/features/enrichment-ui.js
```diff
@@ -0,0 +1,40 @@
+const React = require('react');
```
d2fong: you should move all of your ui components to a singular folder under /features.  Probably in ```features/enrichment``` folder because I think that is the overarching concept.

src/client/router.js
```diff
@@ -45,6 +45,18 @@ module.exports = () => {
           });
           return h(Features.Edit, editProps);
         }
+      },
+      {
+        path: '/gq',
+        render: props => h(Features.GeneQuery, props)
+      },
+      {
+        path: '/enrich',
```
d2fong: Its probably better to have one route defined.  e.g path: '/enrichment-services'.  You can use react tabs to let the testers switch between genequery/enrichment/whatever, which is easier than typing in different urls to do different things.

d2fong: https://github.com/reactjs/react-tabs for usage.  You can also look at some other files in the code base that use react tabs

src/server/routes/pc.js
```diff
@@ -12,6 +13,10 @@ router.get('/querySearch', function (req, res) {
   pc.querySearch(req.query).then(r => res.json(r));
 });
 
+router.get('/gene-query', function (req, res) {
```
d2fong: I think this function should go in /server/routes/rest.js

package.json
```diff
@@ -68,6 +68,7 @@
     "node-fetch": "^1.7.3",
     "object-hash": "^1.2.0",
     "pathway-commons": "^1.5.4",
+    "promise-simple": "^0.1.0",
```
maxkfranz: Why use this instead of native `Promise`s or bluebird?

src/client/router.js
wxli0: two services were merged in /gqenrich

package.json
wxli0: removed, changed to native Promise.

d2fong:
closing old prs

Issue #458, opened 2018-03-01
Interaction view changes


logmosier:
- Add delay to highlight effect/ remove entirely
- Change layout to concentric

jvwong:
Suggestion: Simple way to highlight the query gene in the network? I know it's in the middle but even that sometimes is tricky.

![image](https://user-images.githubusercontent.com/4706307/36988457-e962d558-206c-11e8-91ee-b5cc4aca8589.png)

Issue #457, opened 2018-03-01
Interactions view link


logmosier:
Make the interactions view link button more obvious.

PR #460, opened 2018-03-02
Changes to the interaction view 

logmosier:
Made changes from Issue #457 and #458. The link button is now full width, the layout for the interaction view is now concentric (was grid) and have used debounce on the hover events so the user has to wait to highlight a node/edge.

### DIFF
--- a/src/client/common/cy/events/hover.js
+++ b/src/client/common/cy/events/hover.js
@@ -63,35 +63,37 @@ const baseEdgeHoverStyle = {
   'opacity': 1
 };
 
+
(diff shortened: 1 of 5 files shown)


logmosier:
The changes to the hover node stuff came up during a discussion with @jvwong. How the hover state is it constantly tries to change the state of the the nodes/edges as the user moved the mouse. This leads to a trail of nodes/edges being changed. The change uses _.debounce to prevent the hover style being applied to any of the nodes/edges the user is simply moving the mouse over and waiting till they stop and hover over a node/edge.

d2fong:
Ok.  What is the cancel function doing?

logmosier:
It cancels the debounced function and stops it from happening when the user moves the mouse away form the node/edge.

Issue #475, opened 2018-03-08
gconvert validator issues: add tests and try fetch


wxli0:
On branch gconvert
1. create one basic test in test folder
2. try fetch

wxli0:
didn't get fetch to work, may come back to this issue later

d2fong:
If that is the case, I will close this issue for now.

PR #477, opened 2018-03-09
add a basic test for gconvert validator

wxli0:
relates to #475


jvwong:
Can I suggest 1. Making related issue titles a little more informative 2. Adding some Documentation comments to the empty files?

d2fong:
I think the problem here is that all these empty files should go in a different PR.

@wxli0 this PR has two things:
- adding a bunch of empty files (which is fine, but the PR title says added a test so there is a disconnect between what you are presenting and what you are actually doing)
- added a test for gconvert validator, which is fine too because that is what the PR says to do.

I dont think its useful to add comments to the empty files thats a waste of time.  The empty files are just for folder structure so its better organized.

wxli0:
yea they seem to be dependent on each other. I should have only added the test and then merge it with pr #476 once the review is done.
@jvwong #475 name has been changed

d2fong:
Yeah so do that next time, I am just gonna merge this because we already talked about it, but again this is really important.  At the very least, your PR title should have been something like:

Add folder structure for the enrichment map services, add basic test for gconvert validation

Issue #481, opened 2018-03-09
enrichment service related issues


wxli0:
1. Implement enrichment.js logic (branch enrichment-logic) #482
    - Copy paste from branch emap
2. Encapsulate parsing algorithm in a helper function (branch enrichment-helper) #484 
3. update package.json, package-lock.json (request, csvtojson package)
4. Simplify the process of creating ret json object (branch enrichment-simplify) #484
    - String -> txt -> csv -> object, to string -> array -> object 
5. Try fetch (branch enrichment-fetch)
6. Add basic tests (branch enrichment-test), move onto branch enrichment-test-copy #487 #494 #486
7. Coding style (branch enrichment-style) #492
    - Make default setting a global const 
    - Declare url as const 
    - variable name: str1, str2, str3, str4
8. Expose endpoint (branch enrichment-endpoint) #493
    - router.post
9. missed custbg parameter in enrichment.js defaultSetting (branch enrichment-custbg) #522

wxli0:
@jvwong Which options should be exposed? (output, organism, significant, sortByStructure, orderedQuery, asRanges, noIea, underrep, hierfiltering, userThr, minSetSize, maxSetSize, threadholdAlgo, domainSizeType, custbg, custbgCb)

jvwong:
I would suggest:

- orderedQuery
- userThr 
- minSetSize
- maxSetSize
- threadholdAlgo
- custbg
- custbgCb

You can also expose data sources but default them (see email in scope Google doc).

Also, you should look at the [help](https://biit.cs.ut.ee/gprofiler/help.cgi?help_id=55) and verify the parameters you pasted above even exist (e.g. 'underrep'). You can ask the help desk . I notice you're assuming I know what I'm talking about, which is generally dangerous.

PR #484, opened 2018-03-09
Encapsulate parsing algorithm in a helper function, remove writing files

wxli0:
relates to #481 point2, 4
create helper function
remove writing file

### DIFF
--- a/src/server/enrichment-map/enrichment/enrichment.js
+++ b/src/server/enrichment-map/enrichment/enrichment.js
@@ -13,6 +13,20 @@ const _ = require('lodash');
 const path = require("path");
 
 
+// remove #WARNING and #INFO
+const parseBody = (body) => {
+  // remove the second line
+  const lines = body.split('\n');
+  lines.slice(0, 1);
+  const str1 = body.split('\n').slice(0, 1).join("\n");
+  let str2 = body.split('\n').slice(2).join("\n"); // concatenate at last
+  // remove lines starting with #
+  const str3 = str2.replace(/^#.*$/mg, "");
+  const str4 = (str1 + '\n').concat(str3);
+  return str4;
+}
+
+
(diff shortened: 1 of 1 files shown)


src/server/enrichment-map/enrichment/enrichment.js
```diff
@@ -40,20 +54,13 @@ const enrichment = (query, userSetting) => {
...
-      const str4 = (str1 + '\n').concat(str3);
-      fs.writeFileSync("src/server/enrichment/outputFile", str4);
+
+      fs.writeFileSync("outputFile", parseBody(body));
       // convert csv to json, extract "term id"
       let collection = [];
       let ret = {};
       csv({ delimiter: "\t" })
-        .fromFile("src/server/enrichment/outputFile")
+        .fromFile("outputFile")
```
d2fong: Not sure if this is what we should be doing

wxli0: I changed the pr title. will make a separate issue for it next time.

d2fong:
Nah the title is fine this time.  I was talking about the concept of writing files on request from clients that I am not sure if we should be doing.

wxli0:
that might be related to #481 point4. Change the parsing algorithm in the same way as gconvert so that we can avoid creating the csv file.

src/server/enrichment-map/enrichment/enrichment.js
```diff
@@ -13,6 +13,20 @@ const _ = require('lodash');
 const path = require("path");
 
 
+// remove #WARNING and #INFO
+const parseBody = (body) => {
```
d2fong: use better variable names, e.g:
 parseBody -> parseGProfilerResponse
 body -> gProfilerResponse

These are examples where is is helpful to be less generic

PR #487, opened 2018-03-09
Add a basic test to enrichment service

wxli0:
relates to #481 point 6


d2fong:
Ok great.  This response is much easier to read.

PR #493, opened 2018-03-12
expose an endpoint for enrichment service

wxli0:
relates to #481 point8
request url: 
http://localhost:3000/api/enrichment/?genes=TP53 AFF4
http://localhost:3000/api/enrichment/?genes=TP53&setting={%22custbg%22:%20%22CHD1L%20HDAC3%22}

### DIFF
--- a/src/server/routes/rest.js
+++ b/src/server/routes/rest.js
@@ -67,13 +67,13 @@ router.get('/gene-query', (req, res) => {
 });
 
 //expose a rest endpoint for enrichment
-router.get('/enrichment/', (req, res) => {
+router.get('/enrichment', (req, res) => {
   const genes = req.query.genes;
-  let user_settings = {};
+  let userOptions = {};
   if (req.query.setting !== undefined) {
-    user_settings = JSON.parse(req.query.setting);
+    userOptions = JSON.parse(req.query.options);
   }
-  enrichment(genes, user_settings).then(enrichmentResult => {
+  enrichment(genes, userOptions).then(enrichmentResult => {
     res.json(enrichmentResult);
   });
 });
--- a/src/server/routes/rest.js
+++ b/src/server/routes/rest.js
@@ -67,6 +67,7 @@ router.get('/gene-query', (req, res) => {
 });
 
 //expose a rest endpoint for enrichment
+// use default values if the key is undefined
 router.get('/enrichment', (req, res) => {
   const genes = req.query.genes;
   const tmpOptions = {};
--- a/src/server/routes/rest.js
+++ b/src/server/routes/rest.js
@@ -89,7 +89,7 @@ router.get('/enrichment', (req, res) => {
 
   const userOptions = {};
   for (const key in tmpOptions) {
-    if (tmpOptions[key] != undefined && tmpOptions[key] != '') {
+    if (tmpOptions[key] != undefined) {
       userOptions[key] = tmpOptions[key];
     }
   }
--- a/src/server/routes/rest.js
+++ b/src/server/routes/rest.js
@@ -69,10 +69,30 @@ router.get('/gene-query', (req, res) => {
 //expose a rest endpoint for enrichment
 router.get('/enrichment', (req, res) => {
   const genes = req.query.genes;
-  let userOptions = {};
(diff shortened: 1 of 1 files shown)


src/server/routes/rest.js
```diff
@@ -65,6 +66,18 @@ router.get('/gene-query', (req, res) => {
   });
 });
 
+//expose a rest endpoint for enrichment
+router.get('/enrichment/', (req, res) => {
+  const genes = req.query.genes;
+  let user_settings = {};
+  if (req.query.setting !== undefined) {
+    user_settings = JSON.parse(req.query.setting);
```
d2fong: - use camel case for variables
- how many settings are there?
- generally you should be accessing each setting via it's specific key instead of a an object.

maxkfranz: Could we elide the trailing slash?  It might not make a difference for Express, but it can for some routers and I'd rather keep things simple and consistent.

maxkfranz: Also the names of arguments should be meaningful.  `setting` doesn't mean anything.  It would be better to use something like `options`.

wxli0:
request url: http://localhost:3000/api/enrichment/?genes=TP53%20AFF4
http://localhost:3000/api/enrichment/?genes=TP53 AFF4&custbg=CHD1L HDAC3

parameters: genes, organism, significant, sortByStructure, orderedQuery, asRanges, noIea, underrep, hierfiltering, userThr, minSetSize, maxSetSize, threadholdAlgo, domainSizeType, custbg, custbgCb

src/server/routes/rest.js
```diff
@@ -65,6 +66,39 @@ router.get('/gene-query', (req, res) => {
   });
 });
 
+//expose a rest endpoint for enrichment
+router.get('/enrichment', (req, res) => {
+  const genes = req.query.genes;
+  const tmpOptions = {};
+  tmpOptions.output = req.query.output;
```
d2fong: 1. For each of these options is there some sort of default value?  What does gprofiler do if any of these are undefined?

2. It might also make sense for users to be able to POST with an options JSON body.  It might make more sense than having this as a GET request.

wxli0: 1. Yes. It uses default value when the parameter is undefined. https://github.com/PathwayCommons/app-ui/pull/493/commits/f730ecf54a56c5f8ab08fb234cb76203a4c02ef1#diff-0fd263785891b1a6370048fe62b65e6bL92
https://github.com/wxli0/app-ui/blob/enrichment-endpoint/src/server/enrichment-map/enrichment/enrichment.js#L49

src/server/routes/rest.js
```diff
@@ -65,6 +66,39 @@ router.get('/gene-query', (req, res) => {
   });
 });
 
+//expose a rest endpoint for enrichment
+router.get('/enrichment', (req, res) => {
+  const genes = req.query.genes;
+  const tmpOptions = {};
```
wxli0: add comment saying that gprofiler uses default values when a key is undefined

src/server/routes/rest.js
```diff
@@ -65,6 +66,39 @@ router.get('/gene-query', (req, res) => {
...
+  tmpOptions.userThr = req.query.userThr;
+  tmpOptions.minSetSize = req.query.minSetSize;
+  tmpOptions.maxSetSize = req.query.maxSetSize;
+  tmpOptions.thresholdAlgo = req.query.thresholdAlgo;
+  tmpOptions.domainSizeType = req.query.domainSizeType;
+  tmpOptions.custbg = req.query.custbg;
+  tmpOptions.custbgCb = req.query.custbgCb;
+
+  const userOptions = {};
+  for (const key in tmpOptions) {
```
wxli0: rethink how users send options keys in the request

wxli0:
I think we still need to check 'undefined'. If the user doesn't provide a value for output (http://localhost:3000/api/enrichment/?genes=TP53), tmpOptions.output will be 'undefined', and 'undefined' will override default value 'mini' in [enrichment](https://github.com/wxli0/app-ui/blob/enrichment-endpoint/src/server/enrichment-map/enrichment/enrichment.js#L49).

PR #494, opened 2018-03-12
update test for enrichment service

wxli0:
relates to #481 point6
- update path
- Since object keys are unordered, use a helper to check for object equality

### DIFF
--- a/test/server/enrichment-map/enrichment/enrichment-test.js
+++ b/test/server/enrichment-map/enrichment/enrichment-test.js
@@ -5,7 +5,7 @@ const {enrichment} = require('../../../../src/server/enrichment-map/enrichment')
 
 
 // object keys are unordered
-const objectEquility = (obj1, obj2) => {
+const objectEquality = (obj1, obj2) => {
   _.forEach(obj1, property => {
     if (!obj2.has(property)) { return false; }
     if (obj1.get(property) != obj2.get(property)) { return false; }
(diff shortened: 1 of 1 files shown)


test/server/enrichment-map/enrichment/enrichment-test.js
```diff
@@ -1,72 +1,90 @@
 const chai = require('chai');
 const expect = chai.expect;
-const {enrichment} = require('../../../server/enrichment-map/enrichment');
+const _ = require('lodash');
+const {enrichment} = require('../../../../src/server/enrichment-map/enrichment');
+
+
+// object keys are unordered
+const objectEquality = (obj1, obj2) => {
```
d2fong: lodash should have some function that does this for you.

test/server/enrichment-map/enrichment/enrichment-test.js
```diff
@@ -1,72 +1,90 @@
...
+  _.forEach(obj1, property => {
+    if (!obj2.has(property)) { return false; }
+    if (obj1.get(property) != obj2.get(property)) { return false; }
+  });
+  return true;
+};
+
 
 describe('test enrichment', function() {
+  this.timeout(500000);
```
d2fong: why is this needed

wxli0: Removing this line sometimes gives timeout error.

d2fong: ok

test/server/enrichment-map/enrichment/enrichment-test.js
d2fong: probably see if something like this works:
https://lodash.com/docs/4.17.5#isEqual

wxli0: value for property 't group' is different every time. Ignore the property when checking equality.

PR #495, opened 2018-03-12
Interaction View Tooltips

logmosier:
The changes from  interactionViewCombined that relate to tooltips.
- Getting metadata from the extended SIF file for the tooltip
- Added an option to be passed in to base-network-view to enable tooltips on edges or not
- Tooltips now hide when user clicks off the

### DIFF
--- a/src/client/common/cy/events/show-tooltip.js
+++ b/src/client/common/cy/events/show-tooltip.js
@@ -1,8 +1,9 @@
 
 const MetadataTip = require('../tooltips/');
 
-const bindShowTooltip = (cy) => {
-  cy.on('showTooltip', 'node', function (evt) {
+const bindShowTooltip = (cy,edgeTooltips) => {
+  const selector= 'node'+(edgeTooltips?',edge':'');
+  cy.on('showTooltip', selector, function (evt) {
     const node = evt.target;
 
     const data = node.data();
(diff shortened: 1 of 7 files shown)


src/client/common/cy/events/show-tooltip.js
```diff
@@ -1,8 +1,9 @@
 
 const MetadataTip = require('../tooltips/');
 
-const bindShowTooltip = (cy) => {
-  cy.on('showTooltip', 'node', function (evt) {
+const bindShowTooltip = (cy,edgeTooltips) => {
```
d2fong: Minor nitpick but this should be renamed to communicate that this is a boolean.  It could be interpreted that the param edgeToolTips are actual tooltips being passed in.

src/client/features/interactions/index.js
```diff
@@ -15,7 +15,7 @@ class Interactions extends React.Component {
   constructor(props) {
     super(props);
     this.state = {
-      cy: make_cytoscape({ headless: true, stylesheet: interactionsStylesheet }),
+      cy: make_cytoscape({ headless: true, stylesheet: interactionsStylesheet, edgeTooltips:true }),
```
d2fong: Minor nitpick but this should be renamed to communicate that this is a boolean. It could be interpreted that the param edgeToolTips are actual tooltips being passed in.

PR #496, opened 2018-03-12
fixed bug in enrichment testing

wxli0:
relates to #481 point 4
- enrichment.js, use object instead of map
- enrichment-test.js, 't depth', whitespaces in 't name' also change;

### DIFF
--- a/test/server/enrichment-map/enrichment/enrichment-test.js
+++ b/test/server/enrichment-map/enrichment/enrichment-test.js
@@ -5,8 +5,8 @@ const { enrichment } = require('../../../../src/server/enrichment-map/enrichment
 
 
 // object keys are unordered
-// reponses from gProfiler for 't group', 't depth' and 't name' are different at times
-// ignore 't group', 't name', trim inconsistent whitespaces in 't name'
+// reponses from gProfiler for 'tGroup', 'tDepth' and 'tName' are different at times
+// ignore 'tGroup', 'tDepth', trim inconsistent whitespaces in 'tName'
 const gProfilerResEquality = (obj1, obj2) => {
   return _.isEqualWith(obj1, obj2, (val1, val2, key) => {
     if (key === 'tGroup' || key === 'tDepth') { return true; }
--- a/test/server/enrichment-map/enrichment/enrichment-test.js
+++ b/test/server/enrichment-map/enrichment/enrichment-test.js
@@ -1,90 +1,101 @@
 const chai = require('chai');
 const expect = chai.expect;
 const _ = require('lodash');
-const {enrichment} = require('../../../../src/server/enrichment-map/enrichment');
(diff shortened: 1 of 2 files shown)


test/server/enrichment-map/enrichment/enrichment-test.js
```diff
@@ -1,92 +1,101 @@
...
 
 // object keys are unordered
 // t group changes
 const objectEquality = (obj1, obj2) => {
   return _.isEqualWith(obj1, obj2, (val1, val2, key) => {
-      if (key === "signf" || key === "pvalue" || key === "T" || key === "Q" || key === "Q&T" || key === "Q&T/Q" || key === "Q&T/T" || key === "t type" || key === "t name" || key === "t depth" || key === "Q&T list") {
-        return val1.key === val2.key;
-      }
-      return true;
+    if (key === 't group' || key === 't depth') { return true; }
```
d2fong: Add a comment saying why these keys are ignored specifically

test/server/enrichment-map/enrichment/enrichment-test.js
```diff
@@ -1,90 +1,102 @@
...
+            'Q&T/Q': '1.000',
+            'Q&T/T': '0.012',
+            't type': 'rea',
+            't group': '2',
+            't name': '   RNA Polymerase II Pre-transcription Events',
+            't depth': '1',
+            'Q&T list': 'AFF4'
+          }
+      };
+      expect(objectEquality(result, res)).to.equal(true);
```
maxkfranz: Please read the chai docs:  http://chaijs.com/api/bdd/

`expect(foo).to.deep.equal(bar)`

Don't re-implement things that have been done in libraries.

test/server/enrichment-map/enrichment/enrichment-test.js
```diff
@@ -1,90 +1,102 @@
...
+            T: '83',
+            Q: '1',
+            'Q&T': '1',
+            'Q&T/Q': '1.000',
+            'Q&T/T': '0.012',
+            't type': 'rea',
+            't group': '2',
+            't name': '   RNA Polymerase II Pre-transcription Events',
+            't depth': '1',
+            'Q&T list': 'AFF4'
```
maxkfranz: Is this the expected format of the response?  These are not proper variable names...

wxli0: We only check for specific keys here. renamed the function to 'gProfilerResEquality'.

test/server/enrichment-map/enrichment/enrichment-test.js
wxli0: We only check for specific keys here. renamed the function to 'gProfilerResEquality'.

PR #497, opened 2018-03-12
Interaction View Filter

logmosier:
Adds the filter menu to the interaction view.

### DIFF
--- a/src/client/common/components/base-network-view/base-network-view.js
+++ b/src/client/common/components/base-network-view/base-network-view.js
@@ -126,7 +126,8 @@ class BaseNetworkView extends React.Component {
           onClick: () => {
             btn.func(state);
           },
-          desc: btn.description
+          desc: btn.description,
+          cy: state.cy
         })
       );
     });
@@ -208,7 +209,8 @@ class BaseNetworkView extends React.Component {
       h('div.graph', {
           className: classNames({
             'graph-network-loading': this.state.networkLoading,
-            'graph-sidebar-open': this.state.open
+            'graph-sidebar-open': this.state.open && this.state.activeMenu!='filter-menu',
+            'graph-sidebar-small-open': this.state.activeMenu==='filter-menu'
           }),
         },
         [
@@ -218,7 +220,7 @@ class BaseNetworkView extends React.Component {
         ]
       ),
       h('div', {
-        className: classNames('sidebar-menu', { 'sidebar-menu-open': this.state.open })
+        className: classNames('sidebar-menu',{'sidebar-menu-small':this.state.activeMenu==='filter-menu',  'sidebar-menu-open': this.state.open })
       }, [
           h('div', {
             className: classNames('sidebar-close-button-container', { 'sidebar-close-button-container-open': this.state.open })
--- a/src/client/common/components/base-network-view/base-network-view.js
+++ b/src/client/common/components/base-network-view/base-network-view.js
@@ -105,6 +105,7 @@ class BaseNetworkView extends React.Component {
     const menus = state.componentConfig.menus;
 
     const activeMenu = menus.filter(menu => menu.id === state.activeMenu)[0].func(state);
+    const menuWidth = menus.filter(menu => menu.id === state.activeMenu)[0].width;
 
     const menuButtons = toolbarButtons.filter(btn => btn.type === 'activateMenu').map(btn => {
       return (
@@ -213,9 +214,9 @@ class BaseNetworkView extends React.Component {
       h('div.graph', {
           className: classNames({
             'graph-network-loading': this.state.networkLoading,
-            'graph-sidebar-open': this.state.open && this.state.activeMenu != 'filter-menu',
-            'graph-sidebar-small-open': this.state.activeMenu === 'filter-menu'
+            'graph-sidebar-open': this.state.open
           }),
+          style: { width: menuWidth?`${100-menuWidth}%`:'' }
         },
         [
           h('div.graph-cy', {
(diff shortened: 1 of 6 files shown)


src/client/features/interactions/filter-menu.js
```diff
@@ -0,0 +1,19 @@
+const React = require('react');
+const h = require('react-hyperscript');
+const classNames = require('classnames');
+class InteractionsFilterMenu extends React.Component {
+  render(){
+   const props= this.props;
+   const buttons= [...props.buttons].map(([type, clicked])=>
+   h('div',{key:type,className:classNames ('interaction-filter-button',clicked? 'interaction-filter-clicked':'interaction-filter-not-clicked'),onClick: () => props.filterUpdate(type)},[
```
maxkfranz: Split these up for readability

`'a', 'b', { 'c': condition }`

```
{
  'a': true,
  'b': true,
  'c': condition
}
```

The lines are long. Use newlines on your objects where it makes sense

src/client/features/interactions/index.js
```diff
@@ -26,14 +43,18 @@ class Interactions extends React.Component {
       },
       id:'',
       loading: true,
+      catagories: new Map (),
+      buttons:new Map([['Binding',false],
+        ['Phosphorylation',false],
+        ['Expression',false]
+      ]),
```
maxkfranz: Double arrays are confusing.  What are these values supposed to represent? An object would be clearer.

src/client/features/interactions/index.js
```diff
@@ -44,11 +65,27 @@ class Interactions extends React.Component {
         id: network.id,
         loading: false
       }); 
+      
     });
     this.state.cy.on('trim', () => {
+      const catagories=this.state.catagories;
```
maxkfranz: sp: category

src/client/features/interactions/index.js
```diff
@@ -119,7 +156,30 @@ class Interactions extends React.Component {
...
+  filterUpdate(type) {
+    const state=this.state;
+    const catagories = state.catagories;
+    const buttons=state.buttons;
+    const cy= state.cy;
+    const edges=catagories.get(type).edges;
+    const nodes=catagories.get(type).nodes;
+    
+    hideTooltips(cy);
+    const hovered = cy.filter(ele=>ele.style('background-color')==='blue'||ele.style('line-color')==='orange');
```
maxkfranz: This is wrong.  You're conflating style with state.  Use flags or other real values for meaningful state.  The style should just be mapped from the state using the stylesheet.

src/client/features/interactions/index.js
```diff
@@ -119,7 +156,30 @@ class Interactions extends React.Component {
...
+    removeStyle(cy, hovered, '_hover-style-before');
 
+    if(!buttons.get(type)){
+      cy.remove(edges);
+      cy.remove(nodes.filter(nodes=>nodes.connectedEdges().length<=0));
```
maxkfranz: FYI: There are .empty() and .nonempty() functions if you want

maxkfranz: Also cy.batch() may be useful here

src/client/features/interactions/index.js
maxkfranz: Consider, for example, that the colours may be changed in future and this code would break.

src/client/common/components/base-network-view/base-network-view.js
```diff
@@ -208,7 +213,8 @@ class BaseNetworkView extends React.Component {
       h('div.graph', {
           className: classNames({
             'graph-network-loading': this.state.networkLoading,
-            'graph-sidebar-open': this.state.open
+            'graph-sidebar-open': this.state.open && this.state.activeMenu != 'filter-menu',
```
d2fong: Could you play around with trying to inject the width as a prop?

e.g
```js
h('div.graph', {
  className: classNames(... graph-sidebar-open: this.state.open),
  style: { width: this.state.sidebarWidth }
})
```

d2fong: - [ ] create a default width if no width is specified in the props
- [ ] in your interactions feature, inject the special width into the base view component

src/client/common/components/base-network-view/base-network-view.js
```diff
@@ -218,7 +224,7 @@ class BaseNetworkView extends React.Component {
         ]
       ),
       h('div', {
-        className: classNames('sidebar-menu', { 'sidebar-menu-open': this.state.open })
+        className: classNames('sidebar-menu',{'sidebar-menu-small':this.state.activeMenu === 'filter-menu',  'sidebar-menu-open': this.state.open })
```
d2fong: same here:
- dont use a classname and instead inject the width that you want the menu to be through props

src/client/features/interactions/index.js
```diff
@@ -3,14 +3,32 @@ const h = require('react-hyperscript');
...
+  toolbarButtons: BaseNetworkView.config.toolbarButtons.concat({
+    id: 'filter',
+    icon: 'filter_list',
+    type: 'activateMenu',
+    menuId: filterMenuId,
+    description: 'Filter interaction types'
+  }),
+  menus: BaseNetworkView.config.menus.concat({
+    id: filterMenuId,
+    width: 25, //%
```
d2fong: Does it make sense to have it as a number?  Would it be better to just let it be '25%'?

src/client/common/components/base-network-view/base-network-view.js
```diff
@@ -218,7 +225,8 @@ class BaseNetworkView extends React.Component {
         ]
       ),
       h('div', {
-        className: classNames('sidebar-menu', { 'sidebar-menu-open': this.state.open })
+        className: classNames('sidebar-menu',{'sidebar-menu-open': this.state.open }),
+        style: { width: menuWidth?`${menuWidth}%`:'' }
```
d2fong: What happens to the visuals of the sidebar if width is ''?

logmosier: The width property gets ignored and the width from the class sidebar-menu-open is used

src/client/features/interactions/index.js
logmosier: The value of the width also gets removed from the width of the graph element to make sure it aligns to the edge of the menu still.

Issue #488, opened 2018-03-09
emap service related issues


wxli0:
1. Implement emap logic (branch emap-logic) #499
   - Copy paste from branch emap
2. pathwayTable/preprocess.js, fetchInfo.js (branch emap-preprocess) #506  
   - Coding style
      - Variable name: tmp, data, table
      - Magic number: tmp[0]
3. fetchInfo.js (branch emap-fetchInfo) #523 
     - Remove line 15-17
        - Invalid pathwayids shouldn’t exist in pathwayList when calling fetch function, removed from pathwayList by generateCys
4. generateCys.js (branch emap-generateCys) #507
    - remove make_cytoscape call
    - Coding style
       - let to const: unrecognized, duplicate 
      - variable name: table
       - Magic number: .split(‘_’)[0]...
5. Add basic tests (branch emap-test) #500 
    - in test/enrichment-map/emap/emap-test.js
6. expose endpoint (branch emap-endpoint) #508
     - sample: http://localhost:3000/api/emap/?pathwayIdList=GO:0043525 GO:0043523
7. fix variable name (branch emap-table) #518
     - in src/server/enrichment-map/pathwayTable/index.js table->pathwayInfoTable
8. variable names, const #530
     - fetchInfo.js 
        - fetch -> fetchPathwayInfo
     - intersect.js
        - count-> intersectionCount
        - similarityRate -> pathwayPairGraph
        - similarity -> pathwayListGraph
    - generateCys.js -> generateGraphInfo.js
       - generateCys -> generateGraphInfo
     - let -> const
     - delete fetchInfo.js

PR #507, opened 2018-03-14
remove make_cytoscape call, change variable names in generateCys.js

wxli0:
relates to #488 point 4

### DIFF
--- a/src/server/enrichment-map/emap/generateCys.js
+++ b/src/server/enrichment-map/emap/generateCys.js
@@ -75,5 +75,5 @@ module.exports = { generateCys };
 
 //simple testing
 //console.log(generateCys(["GO:1902275", "GO:2001252", "GO:1905269"]));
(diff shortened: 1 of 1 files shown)


src/server/enrichment-map/emap/generateCys.js
```diff
@@ -46,34 +46,46 @@ const generateCys = (pathwayIdList) => {
     }
   }
   // generate node and edge info
-  let cy = make_cytoscape({ headless: true });
+  const cy = {};
```
wxli0: this should be cytoscapeJSON and not cy

src/server/enrichment-map/emap/generateCys.js
```diff
@@ -46,34 +46,46 @@ const generateCys = (pathwayIdList) => {
...
+  cy.node = [];
```
wxli0: node -> nodes

Issue #501, opened 2018-03-13
Distiller UI


wxli0:
src/client/features/enrichment-map
1. validator.js (on branch validator-ui) #502 
2. enrichment.js (on branch enrichment-ui) #509
3. emap.js (on branch emap-ui) #511 
   - display the network or display json data for constructing the network?

PR #509, opened 2018-03-14
enable UI for enrichment service

wxli0:
relates to #501 point 2

### DIFF
--- a/src/client/features/enrichment-map/enrichment.js
+++ b/src/client/features/enrichment-map/enrichment.js
@@ -6,46 +6,33 @@ class EnrichmentServiceUI extends React.Component {
   constructor(props) {
     super(props);
     this.state = {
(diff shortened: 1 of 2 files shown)


src/client/features/enrichment-map/enrichment.js
```diff
@@ -6,46 +6,33 @@ class EnrichmentServiceUI extends React.Component {
...
-      result_gq: 'no result yet',
-      result_validate_gp: 'no result yet',
-      result_enrich: 'no result yet',
+      enrichResult: 'no result yet',
       query: '',
-      setting: '""'
+      output: '',
+      organism: '',
+      significant: '',
+      sortByStructure: '',
```
d2fong: Are all of these options just strings?

d2fong:
Shift focus from UI to services

PR #511, opened 2018-03-14
enable UI for emap service

wxli0:
relates to #501 point 3


d2fong:
Shift focus from UI to services only

Issue #505, opened 2018-03-13
Expose target, organism for gconvert validator


wxli0:
on branch gconvert-target

wxli0:
wait for #504 to be merged

wxli0:
temporarily working based on local branch gconvert-bug
1. add to endpoint (branch gconvert-target-endpoint)
2. add to ui (branch gconvert-target-ui)

jvwong:
I don't know the answer to this but: 

What if I put in some unrecognized 'target' or 'species' - should we be responsible for telling the client? Or does it just fail silently? I notice gconvert just responds silently (N/A for all genes with 'target=jeffwong'.

wxli0:
#524

PR #519, opened 2018-03-15
exposed target, organism in gconvert validator endpoint

wxli0:
relates to #505


wxli0:
http://localhost:3000/api/gene-query?genes=TP53&target=ENSG

PR #522, opened 2018-03-15
add custbg parameter in enrichment service default options, comment out test

wxli0:
relates to #481 point 9


src/server/enrichment-map/enrichment/enrichment.js
```diff
@@ -39,7 +39,8 @@ const defaultSetting = {
   "max_set_size": "200",
   "threshold_algo": "fdr",
   "domain_size_type": "annotated",
-  "custbg_cb": "none",
+  "custbg": [],
+  "custbg_cb": "0",
```
d2fong: should this be a string or should it be a num

Issue #526, opened 2018-03-16
emap service: exposing options (weighting, cutoff) for edge similarity


wxli0:
1. When calculating similarity rate, expose weight of JC and OC (on branch emap-weight)
    - expose JCWeight, OCWeight=1-JCWeight
    - number between 0 and 1, inclusive
2. Expose cutoff point (on branch emap-cutoff)
    - number between 0 and 1, inclusive

jvwong:
Great. Can we make these issue titles as informative as possible: 'Exposing options (weighting, cutoff) for edge similarity'.

wxli0:
I feel the current pr #529 which only exposes JCWeight might not be flexible for clients.
The following could be better:
1. expose JCWeight, OCWeight
2. when JCWeight is undefined, OCWeight = 1-JCWeight
    when OCWeight is undefined, JCWeight = 1-OCWeight
3. check JCWeight+OCWeight=1 when both are defined. Report errors if not satisfied.

PR #527, opened 2018-03-16
expose cutoff (0.375 by default) from emap endpoint

wxli0:
request url: http://localhost:3000/api/emap/?pathwayIdList=GO:1902275%20GO:1901796&cutoff=-1
report error when cutoff is out of range. 
0.375 by default.
relates to #526 point 2

### DIFF
--- a/src/server/enrichment-map/emap/generateCys.js
+++ b/src/server/enrichment-map/emap/generateCys.js
@@ -22,6 +22,7 @@ const _ = require('lodash');
 // returns a cytoscape object
 const generateCys = (pathwayIdList, cutoff = 0.375) => {
   if (cutoff < 0 || cutoff > 1) { throw new Error('ERROR: cutoff out of range [0,1]');}
+  if (isNaN(Number(cutoff))) { throw new Error('ERROR: cutoff is not a number'); }
   // check unrecognized and duplicates, modify pathwayIdList
   const unrecognized = [];
   const duplicate = [];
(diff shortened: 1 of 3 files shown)


src/server/enrichment-map/emap/generateCys.js
```diff
@@ -20,7 +20,9 @@ const _ = require('lodash');
 
 // input ["GO:1902275", "GO:2001252", "GO:1905269", "GO:0051053"]
 // returns a cytoscape object
-const generateCys = (pathwayIdList) => {
+const generateCys = (pathwayIdList, cutoff) => {
```
d2fong: why not use default params here?  e.g cutoff = 0.375?

src/server/enrichment-map/emap/generateCys.js
```diff
@@ -20,7 +20,9 @@ const _ = require('lodash');
...
+  if (cutoff === undefined) { cutoff = 0.375; }
+  if (cutoff < 0 || cutoff > 1) { return 'ERROR: cutoff out of range [0,1]';}
```
d2fong: You should throw an error instead of returning a string

src/server/enrichment-map/emap/generateCys.js
```diff
@@ -20,7 +20,8 @@ const _ = require('lodash');
 
 // input ["GO:1902275", "GO:2001252", "GO:1905269", "GO:0051053"]
 // returns a cytoscape object
-const generateCys = (pathwayIdList) => {
+const generateCys = (pathwayIdList, cutoff = 0.375) => {
+  if (cutoff < 0 || cutoff > 1) { throw 'ERROR: cutoff out of range [0,1]';}
```
maxkfranz: Do not throw `msg` strings.  Throw `new Error(msg)` objects: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error

wxli0:
should I put the function call in rest.js in try-catch statement?

wxli0:
will make a new pr since the conflicts are too complex to resolve.

PR #529, opened 2018-03-16
expose JCWeight from emap endpoint

wxli0:
request url http://localhost:3000/api/emap/?pathwayIdList=GO:0043525%20GO:0043523&JCWeight=0.9
0.5 by default
report error if out of range
relates to #526 point 2

### DIFF
--- a/src/server/enrichment-map/emap/generateCys.js
+++ b/src/server/enrichment-map/emap/generateCys.js
@@ -20,9 +20,26 @@ const _ = require('lodash');
 
 // input ["GO:1902275", "GO:2001252", "GO:1905269", "GO:0051053"]
 // returns a cytoscape object
-const generateCys = (pathwayIdList, JCWeight) => {
-  if (JCWeight === undefined) { JCWeight = 0.5; }
-  if (JCWeight < 0 || JCWeight > 1) { return 'ERROR: JCWeight out of range [0, 1]'; }
+const generateCys = (pathwayIdList, JCOCWeight) => {
+  let JCWeight = Number(JCOCWeight.JCWeight);
+  let OCWeight = Number(JCOCWeight.OCWeight);
+  if (JCWeight < 0 || JCWeight > 1) {
+    return 'ERROR: JCWeight out of range [0, 1]';
+  }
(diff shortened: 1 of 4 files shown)


src/server/enrichment-map/emap/generateCys.js
```diff
@@ -20,7 +20,24 @@ const _ = require('lodash');
 
 // input ["GO:1902275", "GO:2001252", "GO:1905269", "GO:0051053"]
 // returns a cytoscape object
-const generateCys = (pathwayIdList) => {
+const generateCys = (pathwayIdList, JCOCWeight) => {
+  let JCWeight = Number(JCOCWeight.JCWeight);
+  let OCWeight = Number(JCOCWeight.OCWeight);
+  if (JCWeight < 0 || JCWeight > 1) {
+    return 'ERROR: JCWeight out of range [0, 1]';
```
d2fong: throw errors here instead of returning strings

src/server/enrichment-map/emap/generateCys.js
```diff
@@ -20,7 +20,24 @@ const _ = require('lodash');
...
+  }
+  if (OCWeight < 0 || OCWeight > 1) {
+    return 'ERROR: OCWeight out of range [0, 1]';
```
d2fong: throw errors here instead of returning strings

src/server/enrichment-map/emap/generateCys.js
```diff
@@ -20,7 +20,24 @@ const _ = require('lodash');
...
+  }
+  if (!isNaN(OCWeight) && !isNaN(JCWeight) && OCWeight + JCWeight != 1) {
+    return 'ERROR: OCWeight+JCWeight should be 1';
```
d2fong: throw errors here instead of returning strings

maxkfranz:
Is the emap service doing a calculation or some kind of work?  If it's doing anything more than returning an entry from a db by id, then it should be a post rather than a get.  That means the options can be sent as json -- which means you can send arrays etc.

wxli0:
will make a separate pr for post #546

wxli0:
will make a new pr since the conflicts are too complex to resolve.

Issue #533, opened 2018-03-19
Expose selected parameters for enrichment service


wxli0:
expose
- orderedQuery
- userThr
- minSetSize
- maxSetSize
- threadholdAlgo
- custbg
- custbgCb
on branch enrichment-expose, referring to #481 comments

jvwong:
An excerpt form BIIT support about the data sources. Probably a separate issue? If there is overlap with [Pathway Commons datasources](http://www.pathwaycommons.org/pc2/datasources) (exclude sf_MI, sf_HPA, sf_HP) then we should consider including them. The idea being: What if we swapped out our own enrichment implementation using our own data, then it would be invisible to clients because we are supporting the same data sources.

> In the HTTP POST you can indeed specify only the data sources which you want to keep
>
> The parameters include:
sf_GO  - includes BP, CC, MF subcategories of GO
sf_GO:BP - includes GO biological process terms (if used together with sf_GO then the intersection is applied i.e. sf_GO:BP and sf_GO will give only sf_GO:BP terms)
sf_GO:CC - includes GO cellular component terms
sf_GO:MF - includes GO molecular function terms
sf_KEGG - includes KEGG pathways
sf_REAC - includes Reactome pathways
sf_TF - includes transcription factor predictions from Transfac
sf_MI - includes miRBase predictions
sf_HPA - includes Human Protein Atlas data
sf_CORUM - includes CORUM protein complexes
sf_HP - includes Human Phenotype Ontology terms
sf_BIOGRID - includes Biogrid protein complexes

jvwong:
FYI: [g:Convert option 'underrep'](https://biit.cs.ut.ee/gprofiler/help.cgi?help_id=61)

![image](https://user-images.githubusercontent.com/4706307/37604458-67cb2eae-2b67-11e8-9b6c-90b94ca0b6be.png)

This should default to 0/false and never be exposed.

jvwong:
With regards to parameters `significant`, `threshold`, and `user_thr` there is a relationship between them in g:GOSt that is not easy to discern from their web page. Here's my interpretation.



- `threshold_algo`
  - One of  "analytical", "bonferroni", "fdr"
  - When "fdr" is selected, it [sets the significant threshold to 0.05] (https://biit.cs.ut.ee/gprofiler/help.cgi?help_id=7).    

- `user_thr`
  - Real number in [0, 1].  
[...]

PR #534, opened 2018-03-19
expose selected paramater for enrichment service

wxli0:
relates to #533


wxli0:
will make a new pr after revising

Issue #512, opened 2018-03-14
Reuse enrichment 'gene-validator' service for landing page UniProt lookups


jvwong:
Bear with me...

## Background

To generate the landing page info box for a protein, the search client code [calls](https://github.com/PathwayCommons/app-ui/blob/df90b21a3c175c7a34b084d4d1164b936cf28265/src/client/features/search/index.js#L81) a custom service at [('/uniprotIdSearch')](https://github.com/PathwayCommons/app-ui/blob/df90b21a3c175c7a34b084d4d1164b936cf28265/src/server/routes/pc.js#L16) in an attempt to determine if the user input is associated with a human protein, that is, the code tries to [map an input to a UniProt accession](https://github.com/PathwayCommons/app-ui/blob/df90b21a3c175c7a34b084d4d1164b936cf28265/src/server/pathway-commons/search/index.js#L77):

> app-ui/src/server/pathway-commons/search/index.js 
```js
...
const uniprotIdSearch = async (query) => {
  const queries = await (processPhrase(sanitize(query.q.trim())));
  const filteredQueries = queries.filter(entry=>entry.includes('xrefid'));
  if(!_.isEmpty(filteredQueries)){
    const searchResult = await search()
      .query(query) //input query string
      .q(filteredQueries)
      .format('json')
      .fetch();
    const searchSuccess = searchResult != null
    if (searchSuccess && searchResult.searchHit.length > 0) {
      const filteredResults = searchResult.searchHit.filter(hit =>
        hit.uri.startsWith('http://identifiers.org/uniprot/') 
      );
      return filteredResults.map(hit=>_.last(hit.uri.split('/'))); //Parses and returns the Uniprot id
    }
  } 
  return [];
};
...
```

## Refactoring 

Is it worth considering re-using the ID mapping service at [/gene-query](https://github.com/PathwayCommons/app-ui/blob/df90b21a3c175c7a34b084d4d1164b936cf28265/src/server/routes/rest.js#L60)?

- Benefits
  - Can strip out bespoke 'uniprotid' search code  
  - search logic (see above) in general re-uses [string pattern matching approach](https://github.com/PathwayCommons/pathway-commons-client/blob/master/src/private/constants.js) to determine whether input token is 'ID-like' then queries PC for first 'ProteinReference' with a particular URI pattern, all which is error prone (search 'HGNC:11998').  
[...]

PR #532, opened 2018-03-19
Report invalid 'target', 'organism' in gconvert validator

wxli0:
case insensitive
relates to #524

Issue #524, opened 2018-03-16
Report invalid 'target', 'organism' in gconvert validator


wxli0:
following #505

wxli0:
on branch gconvert-target-validity

jvwong:
Should we declare here, the set of target databases? What about:

- ENSG 
- HGNC (i.e. 'Approved Symbol')
- HGNC_ACC (i.e. 'HGNC ID')
- UNIPROTSWISSPROT (i.e. 'Accession')
- ENTREZGENE (i.e. 'Official symbol')   
- ENTREZGENE_ACC (i.e. 'Gene ID')

There are problems with the validator notions of 'duplicates' etc when we start to open up other database. For example, try 'TP53' in g:Convert with 'target database = REFSEQ_MRNA'

![image](https://user-images.githubusercontent.com/4706307/37526258-3f384288-2905-11e8-9b24-eeec0208f0fc.png)

wxli0:
Are [these](https://github.com/wxli0/app-ui/commit/b75645d11ee4c90771060575314cf8fdbcbf4606#diff-a10c2f2341a355021b49217584c19d14R256) all the target databases allowed?

wxli0:
wait for #519 to be merged

jvwong:
Looks fine to me. In general, we should be forgiving, case insensitive.

wxli0:
case insensitive to both target and organism?

jvwong:
Sure.

jvwong:
Talked to @IgorRodchenkov about 'target' names  the app-ui validator service should use (so that we're not adding to the problem). What about:

| validator parameter | g:Convert parameter name | 
|-------------------- |----------------------|
| ENSG                       | ENSG      | 
| HGNCSYMBOL      | HGNC      |
| HGNC                       | HGNC_ACC    |
| UNIPROT                       | UNIPROT      |
| NCBIGENE                       | ENTREZGENE_ACC      |

This issue references #535 and [pull request 532](https://github.com/PathwayCommons/app-ui/pull/535#discussion_r175883937) by @logmosier.

wxli0:
There is no UNIPROT in g:convert.
![screen shot 2018-03-21 at 9 06 05 am](https://user-images.githubusercontent.com/24982880/37711459-2f08e3be-2ce7-11e8-8202-9b306ef7053b.png)

jvwong:
| validator parameter | g:Convert parameter name | 
[...]

PR #535, opened 2018-03-19
Landing box gene validator

logmosier:
Made changes suggested in issue #512.

### DIFF
--- a/src/client/features/search/index.js
+++ b/src/client/features/search/index.js
@@ -63,12 +63,10 @@ class Search extends React.Component {
   getLandingResult() {
     const state = this.state;
     const query = {
-      q: state.query.q.trim(),
-      type: 'ProteinReference',
-      datasource: state.query.datasource,
-      species: '9606'
+      genes: state.query.q.trim(),
+      target: 'UNIPROTSWISSPROT',
     };
-    if(query.q.includes(' ')){
+    if(query.genes.includes(' ')){
       this.setState({   
         landingLoading: false,
         landing:[]
@@ -78,9 +76,10 @@ class Search extends React.Component {
     this.setState({
       landingLoading: true
     },()=>{
-      ServerAPI.findUniprotId(query).then(res=>{
+      ServerAPI.geneQuery(query).then(res=>{
+        res= res.geneInfo.map(gene=>gene.convertedAlias);
         if(!_.isEmpty(res)){
-          ServerAPI.getProteinInformation(res[0]).then(result=>{
+          ServerAPI.getProteinInformation(res).then(result=>{
             this.setState({
             landingLoading: false,
             landing:result,
--- a/src/client/services/server-api/index.js
+++ b/src/client/services/server-api/index.js
@@ -29,7 +29,7 @@ const ServerAPI = {
   },
 
   geneQuery(query){
-    return fetch(`/api/gene-query?${qs.stringify({genes: query})}`, defaultFetchOpts).then(res => res.json());
+    return fetch(`/api/gene-query?${qs.stringify(query)}`, defaultFetchOpts).then(res => res.json());
   },
 
   findUniprotId(query){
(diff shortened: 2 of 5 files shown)


src/client/features/search/index.js
```diff
@@ -63,12 +63,10 @@ class Search extends React.Component {
   getLandingResult() {
     const state = this.state;
     const query = {
-      q: state.query.q.trim(),
-      type: 'ProteinReference',
-      datasource: state.query.datasource,
-      species: '9606'
+      genes: state.query.q.trim(),
+      target: 'UNIPROTSWISSPROT',
```
maxkfranz: We should probably make all the strings like these defined as enum-like objects in a central file.  Maybe not in this PR, but it should probably be done in a different future PR.

src/client/features/search/index.js
```diff
@@ -78,9 +76,10 @@ class Search extends React.Component {
     this.setState({
       landingLoading: true
     },()=>{
-      ServerAPI.findUniprotId(query).then(res=>{
+      ServerAPI.geneQuery(query).then(res=>{
+        res= res.geneInfo.map(gene=>gene.convertedAlias);
```
maxkfranz: Might want to consider renaming `res` and `result` to clearer names since they're used in the same place.

src/client/features/search/index.js
IgorRodchenkov: What is 'UNIPROTSWISSPROT' and why - there is no such standard name... Why not simply use well-known 'UNIPROT' or any other official synonym for the preferred db name "UniProt Knowledgebase" according to the MIRIAM and Identifiers.org [record](https://www.ebi.ac.uk/miriam/main/datatypes/MIR:00000005)? Why everyone every time tends to invent a new name for known and standard things?

logmosier: UNIPROTSWISSPROT comes from g:convert, they have several different Uniprot target databases so gave each a different name.

jvwong: I knew this would come up :).

This is related to #524. Can you and @wxli0 coordinate to establish the set of 'target' biological database names?

wxli0: The set I'm using is https://github.com/PathwayCommons/app-ui/pull/532/commits/b75645d11ee4c90771060575314cf8fdbcbf4606#diff-a10c2f2341a355021b49217584c19d14R256. Which ones need to be changed?

jvwong: Let's stick to the following set of names, case-insensitive (g:Convert name in parentheses):

1)  'ENSG' ('ENSG') 
2 ) 'HGNC' ('HGNC') 
3)  'HGNC_ID' ('HGNC_ACC')
4)  'UNIPROT' ('UNIPROTSWISSPROT')
5)  'ENTREZGENE' ('ENTREZGENE_ACC')

wxli0: Are 'ENTREZGENE' and 'ENTREZGENE_ACC' the same database?
https://github.com/PathwayCommons/app-ui/issues/524#issuecomment-373731310

IgorRodchenkov: Please, let’s use official db names (ID types) from identifiers.org.
Instead of ‘entrezgene’ (obsolete) or alike, use ‘NCBI Gene’ (e.g., 1234); by ‘HGNC’ we probably here mean ‘HGNC Symbol’ (aka gene name); ‘HGNC’ points to IDs like HGNC:12345 (and these are not widely used nor very useful...)

jvwong: see #524

wxli0:
on pr #532

src/client/features/search/index.js
jvwong: Just to summarize #524:

| validator parameter | g:Convert parameter name | 
|-------------------- |----------------------|
| ENSG                       | ENSG      | 
| HGNCSYMBOL      | HGNC      |
| HGNC                       | HGNC_ACC    |
| UNIPROT                       | UNIPROTSWISSPROT      |
| NCBIGENE                       | ENTREZGENE_ACC      |

logmosier:
If you approve/merge PR #532 I can change the target to be use the new standard database names (UNIPROTSWISSPROT-> UNIPROT).

d2fong:
ok you can change the standard db names now

Issue #536, opened 2018-03-19
Add options in enrichment response


wxli0:
{options: {orderedQuery:…, userThr: …, minSetSize: …, maxSetSize: …, thresholdAlgo: ..., custbg: ..., custbgCb: ...}, pathwayInfo: {<map>}}
on branch enrichment-response

jvwong:
Good catch, I was going to add this to the list as well as for the emap.

PR #538, opened 2018-03-19
add options in enrichment response

wxli0:
relate to #536


src/server/enrichment-map/enrichment/enrichment.js
```diff
@@ -74,12 +74,20 @@ const enrichment = (query, userSetting) => {
       const tNameIndex = 11;
       const tDepthIndex = 12;
       const QTListIndex = 13;
+      ret.orderedQuery = formData.ordered_query;
+      ret.userThr = formData.user_thr;
+      ret.minSetSize = formData.min_set_size;
+      ret.maxSetSize = formData.max_set_size;
```
maxkfranz: The options should not be using snake_case.  Use camelCase

src/server/enrichment-map/enrichment/enrichment.js
```diff
@@ -74,12 +74,20 @@ const enrichment = (query, userSetting) => {
...
+      ret.custbgCb = formData.custbg_cb;
+      ret.pathwayInfo = {};
       _.forEach(responseInfo, elem => {
-        ret[elem[termIdIndex]] = { signf: elem[signfIndex], pvalue: elem[pvalueIndex], T: elem[TIndex], Q: elem[QIndex], tType: elem[tTypeIndex], tGroup: elem[tGroupIndex], tName: elem[tNameIndex], tDepth: elem[tDepthIndex] };
-        ret[elem[termIdIndex]]["Q&T"] = elem[QTIndex];
-        ret[elem[termIdIndex]]["Q&T/Q"] = elem[QTQIndex];
-        ret[elem[termIdIndex]]["Q&T/T"] = elem[QTTIndex];
-        ret[elem[termIdIndex]]["Q&TList"] = elem[QTListIndex];
+        ret.pathwayInfo[elem[termIdIndex]] = { signf: elem[signfIndex], pvalue: elem[pvalueIndex], T: elem[TIndex], Q: elem[QIndex], tType: elem[tTypeIndex], tGroup: elem[tGroupIndex], tName: elem[tNameIndex], tDepth: elem[tDepthIndex] };
+        ret.pathwayInfo[elem[termIdIndex]]["Q&T"] = elem[QTIndex];
```
maxkfranz: Don't repeat yourself https://en.wikipedia.org/wiki/Don%27t_repeat_yourself

e.g. `ret.pathwayInfo[elem[termIdIndex]]`

src/server/enrichment-map/enrichment/enrichment.js
```diff
@@ -91,5 +99,5 @@ const enrichment = (query, userSetting) => {
 module.exports = { enrichment };
 
 // enrichment(['AFF4']).then(function (results) {
-//    console.log(results);
+// console.log(results);
 // });
```
maxkfranz: Commented out code should generally be deleted.

src/server/enrichment-map/enrichment/enrichment.js
wxli0: These are the parameters from [gprofiler help]<https://biit.cs.ut.ee/gprofiler/help.cgi?help_id=55>. 
Should I do something like, const orderedQuery = 'ordered_query'; ret.orderedQuery = formData[orderedQuery]; ?

src/server/enrichment-map/enrichment/enrichment.js
```diff
@@ -74,12 +74,21 @@ const enrichment = (query, userSetting) => {
       const tNameIndex = 11;
       const tDepthIndex = 12;
       const QTListIndex = 13;
+      ret.orderedQuery = formData.ordered_query;
+      ret.userThr = formData.user_thr;
+      ret.minSetSize = formData.min_set_size;
+      ret.maxSetSize = formData.max_set_size;
+      ret.thresholdAlgo = formData.threshold_algo;
+      ret.custbg = formData.custbg;
+      ret.custbgCb = formData.custbg_cb;
```
jvwong: Shouldn't we just remove `custbgCb` as an exposed option and set this (boolean) internally based upon whether `custbg` is null or not?

>custbg_cb - [bool] Use custom statistical background (g:GOSt, g:Cocoa)
custbg - [string] A space-separated list of symbols to use as custom stat. background (g:GOSt, g:Cocoa)

jvwong: In fact, this is the way it works! I entered a background list in custbg and turned the custbgCB to 1 and 0 and it didn't make a difference.

wxli0: That makes sense to me, but...is the option custbgCb useless? Why does it still exist if decisions are made purely based on custbg?

jvwong: I tried a call directly to g:Profiler and custbg_cb didn't make any difference. Remove.

Issue #514, opened 2018-03-14
Reuse enrichment 'gene-validator' service for search 


jvwong:
This issue follows from #512.

## Refactoring

The suggestion is re-use the enrichment service ['/gene-query']() in the [search](https://github.com/PathwayCommons/app-ui/blob/df90b21a3c175c7a34b084d4d1164b936cf28265/src/server/pathway-commons/search/index.js#L12-L31) to test whether a user specified input token is a gene/product. 

> app-ui/src/server/pathway-commons/search/index.js
```js
...
const processPhrase = (phrase) => {
  const sourceList = [
    'uniprot',
    'chebi',
    'smpdb',
    'refseq'
  ];

  const tokens = phrase.toUpperCase().split(/\s+/g);
  
  return getHGNCData(path.join(__dirname,'/hgncSymbols.txt')).then(collection => {
    return tokens.map(token => {
      //if symbol is recognized by at least one source
      const recognized = sourceList.some(source => utilities.sourceCheck(source, token))
                              || collection.has(token.toUpperCase());
      const sanitized = sanitize(token);
      return recognized ? ( 'xrefid:' + sanitized ) : ( 'name:' + '*' + sanitized + '*' );
    });
  });
};
...
``` 

- Benefits 
  -  See #512 
  - Reduces dependency on custom HGNC text file
  - Move away from pathway-commons-client as dependency for ID recognition 

- Caveats
  - Should use original input tokens/strings in any downstream query.
  - Does not support small molecule lookups

logmosier:
Could we still use the source list to recognize the small molecules and then use gene validator for the genes, ie:
```
const sourceList = [
    'smpdb'// and other small molecule sources
  ];
.....
 return geneValidator(query).then(collection => {
   return tokens.map(token => {
      const recognized = was recognized by geneValidator || 
               sourceList.some(source => utilities.sourceCheck(source, token))
```
It would mean we would still be dependent on pathway-commons-client but to a less extant than we are now and we would be able to recognize small molecules .

jvwong:
As far as I understand, the work that [pathway-commons-client](https://github.com/PathwayCommons/pathway-commons-client/blob/356d8d05a8f81dc8b4e13cdd72323103ffa94af5/src/private/constants.js) is doing is pretty much testing against a regex pattern:

>/src/private/constants.js
```js
module.exports = {
  pcAddress: "https://www.pathwaycommons.org/pc2/",

[...]

PR #540, opened 2018-03-19
Search now uses gene validator to recognize genes.

logmosier:
Made the changes discussed in issue #514.

### DIFF
--- a/src/server/pathway-commons/search/index.js
+++ b/src/server/pathway-commons/search/index.js
@@ -1,7 +1,7 @@
 const {search, utilities} = require('pathway-commons');
 const path = require('path');
 const _ = require('lodash');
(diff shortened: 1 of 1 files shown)


src/server/pathway-commons/search/index.js
```diff
@@ -1,7 +1,7 @@
 const {search, utilities} = require('pathway-commons');
 const path = require('path');
 const _ = require('lodash');
-const getHGNCData = require('./hgnc');
+const geneValidator = require('../../enrichment-map/gene-validator/index').validatorGconvert;
```
maxkfranz: Doesn't need to be changed but just FYI: You don't need to specify `index` or `index.js`.  Specifying the dir implicitly refs the index file

PR #543, opened 2018-03-19
Fixed issues that occured from the last PR 

logmosier:
Removed the repeated declaration of interactionsConfig that occured from the most recent PR and combined the componentWillReciveProps.

### DIFF
--- a/src/client/features/interactions/index.js
+++ b/src/client/features/interactions/index.js
@@ -13,15 +13,9 @@ const { BaseNetworkView } = require('../../common/components');
 const { getLayoutConfig } = require('../../common/cy/layout');
 const downloadTypes = require('../../common/config').downloadTypes;
 
-const interactionsConfig={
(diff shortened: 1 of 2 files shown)


src/client/features/interactions/index.js
```diff
@@ -13,15 +13,9 @@ const { BaseNetworkView } = require('../../common/components');
...
-const interactionsConfig={
-  toolbarButtons:_.differenceBy(BaseNetworkView.config.toolbarButtons,[{'id': 'expandCollapse'}],'id'),
-  menus: BaseNetworkView.config.menus,
-  useSearchBar: true
-};  
-
 const filterMenuId='filter-menu';
 const interactionsConfig={
-  toolbarButtons: BaseNetworkView.config.toolbarButtons.concat({
+  toolbarButtons: _.differenceBy(BaseNetworkView.config.toolbarButtons,[{'id': 'expandCollapse'}],'id').concat({
```
maxkfranz: FYI there is also `_.concat()`, which can be better sometimes.  Here it doesn't matter too much

Issue #541, opened 2018-03-19
Expose selected parameters for enrichment service


wxli0:
following #533
on branch enrichment-expose
- orderedQuery
- userThr (default: 0.05)
- minSetSize
- maxSetSize
- thresholdAlgo (default fdr)
- custbg
- custbgCb

1. in rest.js, remove unnecessary expose in get, post request
2. fields following tmpOptions should be in snake case

wxli0:
wait for #538

wxli0:
working on local branch enrichment-backup/enrichment-expose cloned from remote enrichment-response.

PR #545, opened 2018-03-19
expose selected parameters in enrichment service, fixed a bug in rest.js

wxli0:
refer to #541


wxli0:
will make a new pr when #538 is merged.

PR #553, opened 2018-03-21
expose selected parameters for enrichment service, change parameter from camelcase to snakecase

wxli0:
change parameter from camelcase to snakecase since [gprofiler parameters](https://biit.cs.ut.ee/gprofiler/help.cgi?help_id=55) have snakecase naming style.

relates to #541


d2fong:
ok

Issue #565, opened 2018-03-23
Tooltip font size shrinks to unreadable when browser is shortened 


jvwong:
## Environment info

- Browser version (client) : Firefox 59.0.1 (64-bit); Chrome Version 65.0.3325.181 (Official Build) (64-bit); 
- Node.js version (server) : 8.8.1

## Current (buggy) behaviour

When I reduce the length of the browser, the font size of the tooltip title shrinks to the point that it is illegible.

## Desired behaviour

The font size should not change.

## Minimum steps to reproduce

View interactions for [IL4 (P05112)](http://localhost:3000/interactions?ID=P05112), filter by Phosphorylation and view JAK2-IL4 tooltip edge. Then drag browser length smaller.

![bug_shink](https://user-images.githubusercontent.com/4706307/37848799-c5aa7bf0-2eab-11e8-881b-a3e83ce7d514.gif)

PR #569, opened 2018-03-23
Tooltip text shrink fix

logmosier:
Issue #565 fixed.


src/styles/features/view/toolTip.css
```diff
@@ -28,7 +28,7 @@
   top: 2px;
   text-align: center;
   word-wrap: break-word;
-  font-size: 2vh;
+  font-size: calc(12pt+0.3vh);
```
d2fong: does the calculation need to be this complicated?  Can you try using em units?

Issue #585, opened 2018-03-28
Emap service: unrecognized, duplicates


wxli0:
change the logic to
unrecognized -> remove from pathwayIdList, store in unrecognized array
duplicate -> error

on branch emap-error

PR #586, opened 2018-03-28
change duplicate logic in emap service

wxli0:
relates to #585

### DIFF
--- a/src/server/routes/rest.js
+++ b/src/server/routes/rest.js
@@ -35,9 +35,9 @@ router.post('/submit-layout', function (req, res) {
 router.post('/submit-graph', function (req, res) {
   if (isAuthenticated(req.body.token)) {
     controller.submitGraph(req.body.uri, req.body.version, req.body.graph)
-    .then((package) => {
-      res.json(package);
-    });
+      .then((package) => {
+        res.json(package);
+      });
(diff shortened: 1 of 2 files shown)


src/server/routes/rest.js
```diff
@@ -155,7 +155,11 @@ router.post('/enrichment', (req, res) => {
 // Expose a rest endpoint for emap
 router.get('/emap', (req, res) => {
   const pathwayIdList = req.query.pathwayIdList.split(/\s+/);
-  res.json(generateGraphInfo(pathwayIdList));
+  try {
+    res.json(generateGraphInfo(pathwayIdList));
+  } catch (err) {
+    res.json(err.message);
```
d2fong: What is the http status when you return an error message?

wxli0: `info:    GET /api/emap/?pathwayIdList=GO:0043525%20GO:0043523 304 1.533 ms - -`

d2fong: you should probably send a different status.
https://expressjs.com/en/api.html#res.sendStatus

you should probably send a 4XX error, possibly 400, or maybe 500
http://www.restapitutorial.com/httpstatuscodes.html#

Issue #520, opened 2018-03-15
Multi-gene searches


logmosier:
Right now if the user enters multiple genes in the search the landing box does not appear and the user does not have the ability to access the interactions app. A better solution to this situation should be implemented.
I think the landing box could look something like this
![screen shot 2018-03-15 at 12 55 45 pm](https://user-images.githubusercontent.com/10948303/37478305-381a9c86-2850-11e8-8922-d26a1f234024.png)
When the user clicks on more they will be shown the content that would come up from a search of only that gene.

The interactions app could either show the paths between the genes searched or could show the nodes from both of the neighbourhoods interweaved.

Paths between:
pros- would likely be a smaller network,
cons- differs from the behaviour of single gene searches which could be confusing

Multiple neighbourhoods:
pros-in line with current behaviour, not likely to have any unexpected/confusing behaviour for the user 
cons-the network could be extremely large

Also with in the Interactions app I think there should be a way for the user to add new queried genes to the view. This could be added to the filter menu or could be included in its own menu accesed through the toolbar.

jvwong:
Nice! I can see this working. 
One reservation with multiple neighbourhoods is the 'interweave' part - that sounds like 'merging' and could get ugly. I think for pathsbetween you can set a limit:
> limit= [Optional] graph query search distance limit (default = 1).

Your add gene idea is interesting, and starts to sound a lot like [GeneMania](https://genemania.org/) was intended to do.

Also I'm gonna leave this 'big picture' idea, that kinda looks at all of this, including @wxli0 's enrichment services as all different aspects of search...
![unified_search](https://user-images.githubusercontent.com/4706307/37486922-55428470-2866-11e8-8f2c-6c7fd56f0480.png)

logmosier:
Waiting on PR #535

logmosier:
The one problem I can think of is the pathsbetween gives some results that may confuse the user. For instance if you search for ATM and MDM2 which have networks like
![screen shot 2018-03-21 at 12 58 15 pm](https://user-images.githubusercontent.com/10948303/37724806-f9ce16ee-2d07-11e8-93cf-0c2a0a271b53.png)
![screen shot 2018-03-21 at 12 58 09 pm](https://user-images.githubusercontent.com/10948303/37724807-fc487702-2d07-11e8-97c7-5ae3dc1a627e.png)
you get a network like 
[...]

PR #608, opened 2018-04-03
Landing box multi gene

logmosier:
Landing box now handles multiple recognized genes.
Changes made:
- The formatting of the data for the landing box data is now handled mainly in the getLandingResult function instead of in render.
-  handelShowMoreClick added in render to handle the cases in the landing box where the user clicks to reveal more information
- expandableText added in render to handle the function body and synonyms text in the landing box
- CSS changes


jvwong:
Related to #520

Issue #630, opened 2018-04-05
Separate out SIF interaction rules from tooltips


maxkfranz:
Some separate module for handling the rules

Issue #629, opened 2018-04-05
Refactor trim functions


maxkfranz:
e.g. 

https://github.com/PathwayCommons/app-ui/blob/development/src/client/common/cy/tooltips/format-content.js#L33

Issue #624, opened 2018-04-05
Reformat db list as objs


maxkfranz:
https://github.com/PathwayCommons/app-ui/blob/399536631ac23de5a31ce56f5880ad40f2202b54/src/client/common/config.js#L1

e.g. { name: 'Foo', url: 'http://foo.bar', whatever: 'w/e' }

Issue #598, opened 2018-04-02
Interactions Tooltip: Remove 'more' option when only a few 'Links'; Is 'Links' a good description?  


jvwong:
## Description of new feature

The Interations tooltip displays 'Links' which (often?) lists the first PubMed link followed by 'more' option to expand the list. 

1. Should display all in the list when there are just a few (i.e. one link that is hidden).   
2. Do we put anything here other than PubMed links? If not, then 'Links' is unnecesary.

## Motivation for new feature

Reduce unnecesary clicks  by user. Reduce uninformative text.

jvwong:
(Note to @jvwong: Future work on displaying useful publicationxref data).

logmosier:
2. The expression interactions often have both links to pubmed and reactome links

logmosier:
1.  The links in the interaction tooltips currently use the same code as the links in the viewer tooltips. Should this change be applied to the links in the viewer tooltips as well or should it only be applied to the interaction tooltips links.
2. Now that reactome links have been moved under 'Detailed Views' (#613 ) I believe only Pubmed links will appear under links in the interaction tooltips on edges (on nodes there will be other links as well).

jvwong:
> 1. The links in the interaction tooltips currently use the same code as the links in the viewer tooltips. Should this change be applied to the links in the viewer tooltips as well or should it only be applied to the interaction tooltips links.

No - don't alter the SBGN-Viewer tooltip 'Links'. I have this tabled 'elsewhere'.

> 2. Now that reactome links have been moved under 'Detailed Views' (#613 ) I believe only Pubmed links will appear under links in the interaction tooltips on edges (on nodes there will be other links as well).

Let's rename 'Links' as 'PubMed' and stop there. We can log another issue to [e-fetch](https://www.ncbi.nlm.nih.gov/books/NBK25499/#chapter4.EFetch) better citation info.

![image](https://user-images.githubusercontent.com/4706307/38336655-8b39fe58-3830-11e8-9c54-3a08c172f7e5.png)

jvwong:
logged in #620

PR #641, opened 2018-04-05
Tooltip changes

logmosier:
Changes in PR:
- Changed config.databases to an object (issue #624)
- Removed trim functions from fromat-content (issue #629)
- Removed weak link logic from tooltips (kind of issue #630)
- Changed generateDetailedViewList to be more versatile to format links to the viewer or to external links to accomplish issue #598 (formatRenderValue was moved back into generateDatabaseList as it was no longer need for generateDetailedViewList).


src/client/common/cy/tooltips/format-content.js
```diff
@@ -5,54 +5,50 @@ const config = require('../../config');
...
+    let link = db.filter(value => database.toUpperCase() === value.database.toUpperCase());
+    return h('a.db-link' ,{href:'/view?',search: queryString.stringify({
+      uri: link[0].url + link[0].search + data, 
+      title:title, removeInfoMenu:true}),
+    target: '_blank', }, 'Interaction '+(index+1));
+  };
+  const expansionLink = pair[1].length>maxListEntries? h('div.more-link', { onclick: () => expansionFunction(pair[0]) }, trimString(trim)):'';
   if (pair[1].length < 1) { return h('div.error'); }
-  return generateDetailedViewList(sortByDatabaseId(pair[1]), pair[1].length>maxViews, expansionLink,maxViews,title);
+  return interactionList(sortByDatabaseId(pair[1]),trim, expansionLink,maxListEntries,inner);
```
d2fong: can we make the function args consistent i.e make trim always the last param or something like that.

logmosier: I can make it more consistent (trim last almost everywhere) though in the case of line 33 and 103 I think it makes more sense to include title last as it is a parmar that is only used by viwerListHandler so if I were to move trim to last I would need to handle title in all other functions instead of just ignoring it.

Issue #642, opened 2018-04-06
replace 'request' by 'fetch' in validator, enrichment services


wxli0:
https://github.com/PathwayCommons/app-ui/blob/development/src/server/enrichment-map/gene-validator/index.js#L57

https://github.com/PathwayCommons/app-ui/blob/development/src/server/enrichment-map/enrichment/enrichment.js#L80

PR #643, opened 2018-04-06
replace 'request' by 'fetch' in gene validation and enrichment service

wxli0:
related to #642


d2fong:
So fetch is working without any problems?

wxli0:
> So fetch is working without any problems?

seems to be working fine using qs.stringify

Issue #661, opened 2018-04-12
wrong behavior in gene validation service


wxli0:
I noticed something weird in validator service, perhaps a bug on my side (asynchronization).

First set target to 'NCBIGENE', works as expected.
![screen shot 2018-04-12 at 10 15 43 am](https://user-images.githubusercontent.com/24982880/38683331-30581842-3e3b-11e8-9fde-d342716ea663.png)

Then remove parameter target, the service still uses 'NCBIGENE' as target but not 'HGNC'
![screen shot 2018-04-12 at 10 20 59 am](https://user-images.githubusercontent.com/24982880/38683355-3b8d12bc-3e3b-11e8-8981-4eb5a222264f.png)

However, if the service is restarted (ctrl + C, npm run watch), works ok.
![screen shot 2018-04-12 at 10 21 38 am](https://user-images.githubusercontent.com/24982880/38683719-221d972e-3e3c-11e8-9521-8360c8250e1e.png)

wxli0:
Adding {} in _.assign fixes the issue. Can we merge

PR #662, opened 2018-04-12
fix bugs in _.assign

wxli0:
add {} in _.assign
related to #661


wxli0:
@jvwong Can we merge this one?

PR #664, opened 2018-04-12
change file name

wxli0:
@jvwong


jvwong:
Please change file names to all lower case. Separate words with dash.

Issue #666, opened 2018-04-12
Interactions - filter selection states disappear at small widths


jvwong:
When the browser is narrowed, the filter selection states (check, x) disappear. 

This speaks to a related question: Isn't there a more parsimonious way to represent 1. the edge color - interaction mapping and 2. the toggle state of each filter? That is, do I really need a red box AND a green checkmark? For example, if Binding is toggled OFF, do I care that non-existent green edges mean binding?

Breaking case:

[IL4 interactions](http://192.168.81.233:3000/interactions?id=3565&kind=NEIGHBORHOOD)
![image](https://user-images.githubusercontent.com/4706307/38702558-eaa6244a-3e6e-11e8-88af-f18f094f9ac2.png)

PR #679, opened 2018-04-18
Interaction legend rework

logmosier:
@jvwong 
Possible solution to issue #666. Removed checks to indicate a active filter now the legend color only shows when filter is active.

### DIFF
--- a/src/client/features/interactions/filter-menu.js
+++ b/src/client/features/interactions/filter-menu.js
@@ -6,14 +6,14 @@ const _ = require('lodash');
 class InteractionsFilterMenu extends React.Component {
   render(){
     const props= this.props;
-    const buttons= _.map(props.buttonsClicked,(clicked,type)=>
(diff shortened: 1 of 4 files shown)


src/client/features/interactions/filter-menu.js
```diff
@@ -13,9 +13,8 @@ class InteractionsFilterMenu extends React.Component {
         onClick: () => props.filterUpdate(type)
       },
       [
-        h('div',{className:classNames(type,'interaction-filter-legend')}),
+        h('div',{className:classNames('interaction-filter-legend',{[type]:!clicked})}),
```
d2fong: what does ```[type]: !clicked``` mean?

logmosier: The [type] is the class that will be applied based off of which button it is and  [type]: !clicked only applies that class if the filter is active.

Issue #688, opened 2018-04-19
Search page: Clicking browser back/forward buttons change URL but not view 


jvwong:
Clicking the forward or back buttons in the pathway search page results in the location text changing, however, the view does not change.

The expectation is that the buttons should change the URL which should be reflected in the view.

Reproduce this behaviour:

1. Search for a gene like [TP53](http://apps.pathwaycommons.org/search?gt=0&lt=250&q=TP53&type=Pathway) 
2. Search for another gene like [BGN](http://apps.pathwaycommons.org/search?gt=0&lt=250&q=BGN&type=Pathway) 
3. Hit back button. Still shows results for BGN.

PR #752, opened 2018-05-10
Clicking browser back/forward buttons change URL but not view

q4wen:
Clicking the forward or back buttons in the pathway search page now results in both of the URL and the view changing.

Refs: #688


src/client/features/search/index.js
```diff
@@ -33,13 +33,15 @@ class Search extends React.Component {
...
-    
+
       ServerAPI.datasources()
         .then(result => {
           this.setState({
             dataSources: Object.values(result)
           });
         });
+
+    window.onpopstate = (() => window.location.reload() );
```
maxkfranz: This workaround may function correctly, but it might not be the proper solution.  I suspect there are places in the code that aren't using the react-router Link component.

maxkfranz:
You can search for `h('a'` to find plain anchors that should probably be converted.

q4wen:
I tried replacing <button> with <a>, but the page still does not reload on back button clicked.

screen shot 2018-05-14 at 4 30 05 pm

maxkfranz:
You must use the Link component for the router history to work: https://reacttraining.com/react-router/web/api/Link

q4wen:
@maxkfranz @d2fong I implement the `<Link>` component to solve the back/forward button issue, but I am not sure if it is good practice.

Issue #733, opened 2018-05-03
Infobox description: Truncate by sentence; Button text updates 


jvwong:
## Description of new feature

1. Truncate the descriptive text after the conclusion of a sentence rather than by character count or length. 

2. The text  'show' and 'hide' should be updated to 'more' and 'less', respectively.

## Motivation for new feature

Truncating by sentence provides complete statements without further clicks. Updating button text better reflects the consequences of user action.

![image](https://user-images.githubusercontent.com/4706307/39557830-7274fca4-4e58-11e8-9c79-dc5156098145.png)

jvwong:
@q4wen Is the impementation intended to show a certain number of sentences? Period-delimited? The reason I ask is that I am seeing examples where only the first sentence is displayed (e.g. search [ATM](http://appsbeta.pathwaycommons.org/search?gt=0&lt=250&q=ATM&type=Pathway)):

![image](https://user-images.githubusercontent.com/4706307/40003805-4aa10976-5762-11e8-8c6a-e68de17ffb6f.png)

![image](https://user-images.githubusercontent.com/4706307/40003889-7730c576-5762-11e8-900c-d3ec717d59df.png)

jvwong:
Sure, let's try to be consistent in the number of lines (say 2) then move on.

PR #758, opened 2018-05-14
Infobox description more-less style changes

q4wen:
Info box description is now always truncated to 2 lines initially. ref: #733

### DIFF
--- a/src/client/features/search/landing-box.js
+++ b/src/client/features/search/landing-box.js
@@ -165,6 +165,15 @@ const expandableText = (controller,landing,length,text,separator,type,cssClass,t
   return result;
 };
 
+const expandableFunctionText = (controller,landing,text,toggleVar,index)=>{
+  let result = null;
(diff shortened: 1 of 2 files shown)


maxkfranz:
Would you attach a before-and-after screenshot, please?

q4wen:
Before:
before_less


before_more
<br/><br/><br/>
After:
screen shot 2018-05-15 at 9 13 53 am
<br/>
screen shot 2018-05-15 at 9 14 16 am

maxkfranz:
@jvwong @d2fong How does this look to you?

I think the feature itself looks fine.  Maybe there could be a margin around the description paragraph to give a bit more spacing -- but that's pretty minor.

d2fong:
I think it is better now that it is using css instead of some hardcoded values in js.  Improving the styling / refactoring this probably needs to be its own ticket.

Issue #764, opened 2018-05-16
Require access to Pathway Commons URI via environment variable.


jvwong:
## Description of new feature

The URI for Pathway Commons should be declared and accessed in one place, through an environment variable. 

## Motivation for new feature

This will enable the app to fetch from any valid host/domain and ease maintenance.

IgorRodchenkov:
@jvwong @d2fong @maxkfranz @MichaelWrana 

There are many places (a dozen or so) in the apps code where either `www.pathwaycommons.org` or `pathwaycommons.org` or `www.pathwaycommons.org/pc2/` is hardcoded as string.
However, there are several different PC servers, such as production PC9  (www.pathwaycommons.org/pc2/), beta (beta.pathwaycommons.org/pc2/), experimental  (textmining.pathwaycommons.org/pc2/), etc.

Therefore, tt least the PC web **service** base URL should be configurable via an enviroment variable, e.g., PC_URL - just like PORT and NODE_ENV (and also - described in the README). 

Probably, Pathway Commons project home page URL http://www.pathwaycommons.org, which is probably used in links to logo or html pages, does not need to be configurable (can be simply a global constant defined just once; note that simply adding '/pc2/' to the PC home URL, although works right now, won't necessarily resolve to the PC web service endpoint URL in the future), unlike the PC service endpoint URL. 

(I just wanted to try the latest search/view app connect to the beta PC10 web services, and could not do without refactoring/patching apps-ui code...)

d2fong:
https://github.com/PathwayCommons/pathway-commons-client/pull/8

IgorRodchenkov:
Done.

PR #766, opened 2018-05-17
Removed Plaintext Instances of pathway commons URL & URI

MichaelWrana:
References Issue #764 
Replaced with an environment variable, which can be easily modified in `src/config.js` by replacing the following values:
code_image

BASE_URL has now replaced instances of:
`http://www.pathwaycommons.org`
`https://www.pathwaycommons.org`
`http://pathwaycommons.org`
`https://pathwaycommons.org`

PC_URI has now replaced instances of:
`http://www.pathwaycommons.org/pc2/`
`https://www.pathwaycommons.org/pc2/`
`http://pathwaycommons.org/pc2/`
`https://pathwaycommons.org/pc2/`

There are still some instances remaining.  Specifically in the files:
`src/scripts/datasources.js`
`src/client/features/paint/demo-pathway-results.json`
`src/server/graph-generation/generic-physical-entities/generic-physical-entity-map.json`
This is due to the volume of replacements necessary, some of these files have over 10000 instances.

### DIFF
--- a/src/client/common/components/base-network-view/base-network-view.js
+++ b/src/client/common/components/base-network-view/base-network-view.js
@@ -9,6 +9,9 @@ const IconButton = require('../icon-button');
 
 const debouncedSearchNodes = _.debounce(require('../../cy/match-style'), 300);
 
+const conf = require("../../config");
+const BASE_URL = conf.BASE_URL;
+
 // cytoscape
 // grapjson
 // metadata
@@ -178,7 +181,7 @@ class BaseNetworkView extends React.Component {
       h('div', { className: classNames('menu-bar', { 'menu-bar-margin': state.activeMenu }) }, [
         h('div.menu-bar-inner-container', [
           h('div.pc-logo-container', [
-            h('a', { href: 'http://www.pathwaycommons.org/' } , [
+            h('a', { href: BASE_URL } , [
               h('img', {
                 src: '/img/icon.png'
               })
--- a/src/client/common/config.js
+++ b/src/client/common/config.js
@@ -1,3 +1,6 @@
+const conf = require("../../config");
+const PC_URI = conf.PC_URI;
+
 const databases = [
   {database:'BioGrid', url:'http://identifiers.org/biogrid/', search:''},
   {database:'DrugBank', url:'https://www.drugbank.ca/', search:''},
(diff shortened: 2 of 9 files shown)


maxkfranz:
If there are so many instances per file, for those files you mentioned, it may make sense to have a rule to build the URLs from those files.  That would reduce repetition.

d2fong:
That last file contains the most instances and that is a file generated by pathway commons to find gene synonyms used by the painter app.

maxkfranz:
Could we delete the URLs  in that generated file as a post-processing step and generate the URLs on the fly?

d2fong:
We could, but it might not just be the URLS that we need to replace.  It might need to be the whole file itself because it is specific to the version of pathway commons.

maxkfranz:
I see.  Maybe it would be best to tackle those files in a separate issue

IgorRodchenkov:
There is no need for BASE_URL environment variable, for PC home page will be always www.pathwaycommons.org, and this app can be run anywhere (app.pathw..., appsbeta.pathway..., www.pathway.../apps/, etc.. and this is not its base URL anyway)

IgorRodchenkov:
Won't merge.

IgorRodchenkov:
OMG. Dylan is correct. @d2fong @maxkfranz @jvwong 
So, `http://pathwaycommons.org/pc2/` (and probably **misspelled** `https://pathwaycommons.org/pc2/` - should **not** be `https`) in all those test and data json (jsonld) files, such as _generic-physical-entity-map.json_ (as well as in all biopax, sbgn, extended sif) **has nothing to do with web service endpoint** location. This is the BioPAX (RDF/XML) model's _xml:base_, part of some Pathway Commons BioPAX objects' URIs; it looks like URL but in fact is about identity and to be used with pc _/get_ and _/graph_ query parameters like _source, uri, target_. These are **DATA** not links. We should try not to hard-code these. Why we have those data files in the source code, in src (not even resources)? 
What if new PC10 beta or production web services and data is released tomorrow; are we going to patch apps code via find/replace? It's already running at beta.pathwaycommons.org/pc2/ (there is even a different instance at textmining.pathwaycommons.org/pc2/) , and we shall be able to relatively easily switch the apps between production and beta, etc.

d2fong:
This data file is needed by the paint app to get gene synonyms to find as many nodes as possible to map onto them.  That file is a filtered/processed version of the ~179mb file physical_entities.json file you made a few months ago @IgorRodchenkov.  This filtered version needs to be built everytime the data in PC changes i.e every PC release.  We could probably make it a build step when the app is built.

Issue #767, opened 2018-05-17
geneQuery POST uses invalid Content-Type form-urlencoded


jvwong:
## Environment info

- Browser version (client) : n/a
- Node.js version (server) : 8

## Current (buggy) behaviour

In several instances, the search infobox code calls [geneQuery](https://github.com/PathwayCommons/app-ui/blob/development/src/client/services/server-api/index.js#L31-L41) using POST with Content-Type (application/x-www-form-urlencoded).   

> /src/client/services/server-api/index.js
```js
  geneQuery(query){
    query.genes=_.concat(['padding'],query.genes.split(' '));
    return fetch('/api/validation', {
      method:'POST', 
      headers: {
        'Accept': 'application/json',
        'Content-Type': 'application/x-www-form-urlencoded'
      },
      body:qs.stringify(query)
    }).then(res => res.json()).then(ids=> _.assign(ids,{unrecognized:_.tail(ids.unrecognized)}));//remove padding
  },

```

Why this still works is not completely clear to me but it is accepted by the service if I add a `gene` value `padding`:

![image](https://user-images.githubusercontent.com/4706307/40204480-bb22ae34-59f6-11e8-8201-c5d182214a20.png)

## Desired behaviour

POST data with Content-Type application/json

q4wen:
Not sure I change correctly.

There is 400 bad request error when I change the content type to application/json
![screen shot 2018-05-18 at 1 24 27 pm](https://user-images.githubusercontent.com/16858225/40250157-66fd2372-5aa3-11e8-8a5e-0c1958e6107a.png)
<br/><br/>
![screen shot 2018-05-18 at 1 24 44 pm](https://user-images.githubusercontent.com/16858225/40250159-68e3b4ee-5aa3-11e8-96b4-8b68db142c25.png)

And here is the related back-end code:
![screen shot 2018-05-18 at 1 58 57 pm](https://user-images.githubusercontent.com/16858225/40250224-a5ad347c-5aa3-11e8-8906-58c8a5ec970f.png)

I tried adding `res.setHeader('content-type', 'application/json')`, but still have the same error.

jvwong:
Service would be expecting JSON with your change - but sending some string?

```js
geneQuery(query){
...
      body:qs.stringify(query)
   ...
```

maxkfranz:
The body is usually either FormData or stringified JSON.  Example: https://stackoverflow.com/questions/29775797/fetch-post-json-data

[...]

PR #774, opened 2018-05-22
Correct content-type in geneQuery

q4wen:
Content type is changed to `application/json`
Request body is changed to JSON format.

ref: #767


d2fong:
@jvwong can you confirm that this is still what you want per issue #767?

q4wen:
Yes. Tested in local server. @jvwong

Issue #761, opened 2018-05-15
Remove extra steps to Datasource   


jvwong:
## Description of new feature

Related to #687 

We have a 'Datasource' toggle that shows a dropdown labelled with the current datasource. The user needs to look across the the screen to the datasource label, then select from the dropdown. 

![image](https://user-images.githubusercontent.com/4706307/40060551-e42a5b66-5824-11e8-8d63-3aac926e270c.png)

Question: Is there a simpler way to offer this?

## Motivation for new feature

Reduce the amount of work needed by a user to filter.

jvwong:
Maybe ditch the 'Datasource' toggle in your second image?

CaitlinOCallaghan:
So remove the 'Datasource' header and have the filter drop down appear upon loading?

jvwong:
Yes. Simple dropdown to select datasource.

CaitlinOCallaghan:
Here is what it looks like on full screen:
![image](https://user-images.githubusercontent.com/38890251/40073590-3edc3e7e-5845-11e8-977e-544bdcb7797c.png)

If you decrease screen width, the filter menu slides with the rest of the search page so that it is always at the edge of the description box. It also fits on phone views. My only concern is that when you open the menu on the phone, the display is:
![image](https://user-images.githubusercontent.com/38890251/40073774-d171ee00-5845-11e8-93bc-913ec38da42c.png)

Does this mean the filter menu will be slightly cut off when opened on phone?

jvwong:
I think these things always depends on the phone/browser. I could live without filtering on small devices. What do you think?

CaitlinOCallaghan:
I agree, because the filter bar does show up on the screen; however, it's when you open it that a portion may be cut off on smaller phones. 

![image](https://user-images.githubusercontent.com/38890251/40074625-16129634-5848-11e8-88fa-1194837fb772.png)

CaitlinOCallaghan:
Are there any more changes that I should make, or can I submit a pull request?

jvwong:
Let's go with this.

maxkfranz:
The dropdown only looks like that in the Chrome debug mode in the browser.  It doesn't get all of the details right when it's in the "mobile phone simulator" mode.  If you use an iPhone, you'll get the rolodex UI for the select box.

IgorRodchenkov:
@jvwong  @CaitlinOCallaghan 

[...]

PR #778, opened 2018-05-22
Issue 764 etc

IgorRodchenkov:
Implemented PC_URL environment option made other improvements, such as simplified PC web service queries - pathway-commons-client (npm) module is not required anymore, and the pc client code is now simpler and reused better.

Also implemented one solution for #761 (see my comments there).


IgorRodchenkov:
Fixes #764

src/client/features/interactions/index.js
```diff
@@ -51,12 +52,18 @@ class Interactions extends React.Component {
...
 
     const query = queryString.parse(props.location.search);
-    const sources=_.uniq(_.concat([],query.source));
-    const kind = sources.length>1?'PATHSBETWEEN':'Neighborhood';
-    ServerAPI.getNeighborhood(sources,kind).then(res=>{
+    const sources = _.uniq(_.concat([],query.source)); //IDs or URIs
+    const params = {
+      source : sources,
+      pattern : ['controls-phosphorylation-of','in-complex-with','controls-expression-of', 'interacts-with'],
+      kind : sources.length>1 ? 'pathsbetween' : 'neighborhood',
```
jvwong: Is it the case that 'pathsbetween' will not return 'interacts-with'?

IgorRodchenkov:
Ah, let me fix a bug in pc queries to download pathway data in different formats (SIF, BioPAX, etc.) from the right panel...

IgorRodchenkov:
Let me fix a refactoring bug first - related to /pc-client/ routes, queries for data to download..

But, no, all these changes have nothing to do with the other issue about and pathsbetween, which never finds interacts-with due its implementation in java (Paxtools/pattern).

IR

src/client/features/search/index.js
```diff
@@ -34,12 +34,12 @@ class Search extends React.Component {
...
-      ServerAPI.datasources()
-        .then(result => {
-          this.setState({
-            dataSources: Object.values(result)
-          });
-        });
+    ServerAPI.datasources()
+    .then(result => {
+      this.setState({
+        dataSources: Object.values(result).filter(ds => ds.hasPathways==true)
```
jvwong: `hasPathways` new in PC v10?

src/client/features/interactions/index.js
```diff
@@ -66,12 +73,17 @@ class Interactions extends React.Component {
...
       });
     });
-    //get ids from uris
-    const geneIds = sources.map(source=>
-      source.includes('pathwaycommons')?
-      ServerAPI.pcQuery('traverse',{uri:source,path:`${_.last(source.split('/')).split('_')[0]}/displayName`}).then(result=>result.json())
-      .then(id=> _.words(id.traverseEntry[0].value[0]).length===1 ? id.traverseEntry[0].value[0].split('_')[0] : ''):
-      source.replace(/\//g,' ')
+    //get ids from uris 
+    const geneIds = sources.map(source =>
```
jvwong: What case was this traverse trying to handle?

src/client/features/interactions/index.js
IgorRodchenkov: It says "get ids from uris" in the comment line above that code (if, e.g., an uri is standard PR's uri, I bet); @logmosier can comment on exact purpose of this. I did not change the purpose of this code. It would be great to have more comments in the code, in many places.

My changes (using one universal new method `ServerAPI.pcQuery` instead of `ServerAPI.getNeighborhood`): https://github.com/PathwayCommons/app-ui/commit/bc43b132985cc406cb9502ac9165ee73cd8a0d69#diff-48d6e306661f0412cd8851fd566e8bba
`ServerAPI.getNeighborhood` (removed by https://github.com/PathwayCommons/app-ui/commit/bc43b132985cc406cb9502ac9165ee73cd8a0d69#diff-0e447967aaeb299d989022e921439ad2) used the same graph query parameters.

PS (off-topic):
Also, look how I filter datasources for that drop-down menu (to keep only those actually having pathways): https://github.com/PathwayCommons/app-ui/commit/9b745adb006efe2f6ae8e81a2d82f10334e8250f#diff-0e447967aaeb299d989022e921439ad2

src/client/features/search/index.js
IgorRodchenkov: No, it was always returned for each datasource in the array - by  /pc2/metadata/datasources query.
I just added this filed to the server-side pc code (https://github.com/PathwayCommons/app-ui/blob/issue_764_etc/src/server/pathway-commons/index.js#L107) and can filter as shown above.

PS:
note that internal pc client module is now just one file: src/server/pathway-commons/index.js, and there is no 'pathway-commons' import from npm.

src/client/features/interactions/index.js
IgorRodchenkov: It says "get ids from uris" in the comment line above that code (if, e.g., an uri is standard PR's uri, I bet); @logmosier can comment on exact purpose of this. I did not change the purpose of this code. It would be great to have more comments in the code, in many places.

IgorRodchenkov:
I am curious, who named a branch 'garbage-pc-search'? So untrue and hurts for some reason :/

Issue #790, opened 2018-05-29
ServerAPI method querySearch mangles searches terminating with <db>:<id> 


jvwong:
## Current (buggy) behaviour

 
We allow query of the format `q=db:id` to signify referral from biological database. A requirement is that the pathway search sent to PC web service uses only the `id` portion. To this end, the [`querySearch`](https://github.com/PathwayCommons/app-ui/blob/62d497192378d9614df05db5041bfbcf7bda6479/src/client/services/server-api/index.js#L29) tests for this pattern and strips out the `db:` prefix:

> /src/client/services/server-api/index.js#L29
```js

const ServerAPI = {
...
  querySearch(query){
    const queryClone=_.assign({},query);
    if(/(uniprot:\w+|ncbi:[0-9]+|hgnc:\w+)$/.test(queryClone.q)){queryClone.q=queryClone.q.split(':')[1];}
    return fetch(`/pc-client/querySearch?${qs.stringify(queryClone)}`, defaultFetchOpts).then(res => res.json());
  },
...
}
```   
Edge case (rare) is  that when multiple tokens are entered, with the final one being of the format `db:id` then entire query including and before the `db:` prefix is removed.

Example case: Search for [`MDM2 ncbi:7157 `](http://appsbeta.pathwaycommons.org/search?gt=0&lt=250&q=MDM2%20ncbi%3A7157%20&type=Pathway) results in a PC search with `q=7157`.   

## Desired behaviour

The search should ideally send the entire search phrase as is

IgorRodchenkov:
I think, we introduced db:id queries as a special case for linkout/cross-reference from UniProt, GeneCards, NCBI. Ideally, we should not allow multiple terms in such queries. Jeff is right - typing db:id plus other words in the text input field needs to be processed as it used to be before introducing db:id feature...
Good catch!

maxkfranz:
Wouldn't it make sense that ehe `db:id` format should be valid to use everywhere in the app-ui code -- except just before passing the ID to external PC services?

PR #794, opened 2018-05-30
Text input field allows multiple tokens

q4wen:
Text input field accepts multiple tokens of the form `db:id` (ref:#790):
<br/>
screen shot 2018-05-30 at 10 47 34 am
<br/>
The procedure is:

1. Remove all prefix db:id in input string
2. Remove extra spaces
3. Search ids as before.

### DIFF
--- a/src/client/services/server-api/index.js
+++ b/src/client/services/server-api/index.js
@@ -26,7 +26,7 @@ const ServerAPI = {
 
   querySearch(query){
     const queryClone=_.assign({},query);
-    queryClone.q=queryClone.q.replace(/(uniprot|ncbi|hgnc)\s*:/gi,"").replace(/\s\s+/g, ' ').replace(/\s+$/, '');
+    queryClone.q=queryClone.q.replace(/^(uniprot|ncbi|hgnc):/i,"");
     return fetch(`/pc-client/querySearch?${qs.stringify(queryClone)}`, defaultFetchOpts).then(res => res.json());
   },
 
--- a/src/client/services/server-api/index.js
+++ b/src/client/services/server-api/index.js
@@ -26,7 +26,9 @@ const ServerAPI = {
 
   querySearch(query){
     const queryClone=_.assign({},query);
-    queryClone.q=queryClone.q.replace(/^(uniprot|ncbi|hgnc):/i,"");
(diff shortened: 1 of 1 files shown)


src/client/services/server-api/index.js
```diff
@@ -26,7 +26,7 @@ const ServerAPI = {
 
   querySearch(query){
     const queryClone=_.assign({},query);
-    if(/(uniprot:\w+|ncbi:[0-9]+|hgnc:\w+)$/.test(queryClone.q)){queryClone.q=queryClone.q.split(':')[1];}
+    queryClone.q=queryClone.q.replace(/(uniprot|ncbi|hgnc)\s*:/gi,"").replace(/\s\s+/g, ' ').replace(/\s+$/, '');
```
jvwong: Can we trigger the replace only when there is a single token ? Maybe:

`q=queryClone.q.replace(/^(uniprot|ncbi|hgnc):\w+$/i,"")...`

src/client/services/server-api/index.js
```diff
@@ -26,7 +26,7 @@ const ServerAPI = {
 
   querySearch(query){
     const queryClone=_.assign({},query);
-    if(/(uniprot:\w+|ncbi:[0-9]+|hgnc:\w+)$/.test(queryClone.q)){queryClone.q=queryClone.q.split(':')[1];}
+    queryClone.q=queryClone.q.replace(/^(uniprot|ncbi|hgnc):/i,"");
```
jvwong: Correct me if I am wrong but won't this just replace the first token?

q4wen: Yes, it will only remove db when the query has format db:id (ncbi:7156 => 7156); otherwise, don't do anything. Maybe I should restrict the condition so that "ncbi:7156 mdm2" won't be replaced.

Should I also make validation of database name (ncbi, uniprot, and hgnc) case sensitive since the links we provide for Unprot should be all lower case?

jvwong: >  Maybe I should restrict the condition so that "ncbi:7156 mdm2" won't be replaced.

Yes.

> Should I also make validation of database name (ncbi, uniprot, and hgnc) case sensitive since the links we provide for Unprot should be all lower case?

No.

Issue #805, opened 2018-06-01
App hangs and throws error upon slow/failed response from g:Profiler


jvwong:
## Environment info

- Browser version (client) : n/a
- Node.js version (server) : 8

## Current (buggy) behaviour

The search is non-responsive (spinner) when trying to search for any term. This coincides with non-responsive calls to g:Profiler's  g:Convert.

![image](https://user-images.githubusercontent.com/4706307/40815306-89d7be82-6513-11e8-947b-8a6b40a3f691.png)

The server console outputs the following:

```js
(node:10781) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without acatch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 11)
(node:10781) UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'substring' of undefined
    at cleanUpEntrez (/Users/jeffreywong/Projects/PathwayCommons/web/app-ui/src/server/enrichment/helper/clean-up-entrez.js:5:20)
    at _.forEach.info (/Users/jeffreywong/Projects/PathwayCommons/web/app-ui/src/server/enrichment/validation/index.js:78:26)
    at arrayEach (/Users/jeffreywong/Projects/PathwayCommons/web/app-ui/node_modules/lodash/lodash.js:508:11)
    at Function.forEach (/Users/jeffreywong/Projects/PathwayCommons/web/app-ui/node_modules/lodash/lodash.js:9334:14)
    at fetch.then.then.body (/Users/jeffreywong/Projects/PathwayCommons/web/app-ui/src/server/enrichment/validation/index.js:75:11)
    at <anonymous>
    at process._tickCallback (internal/process/next_tick.js:188:7)
```

## Desired behaviour

Should ideally allow failure of dependencies gracefully and enable return of PC web service results. 

## Minimum steps to reproduce

Running search locally. Search by clicking sample search.

jvwong:
@CaitlinOCallaghan let's both take a stab at this. Idea is the `validator` should honour its contract to clients even in the case where the underlying dependency (i.e. g:Convert)  1. is unreasonably slow to respond or 2. fails. My fault for not looking into all these Promise/fetch/catch...

Tasks
 - [ ]   TODO.

maxkfranz:
For now, could we revert the problematic commit?  The bugs can be addressed in the feature branch and a new PR can be made once everything's working.

jvwong:
[...]

PR #807, opened 2018-06-04
Iss805 validator hang

jvwong:
Related to #805 

- For node-fetch 
  - Added a timeout (5s) that thresholds round-trip to g:Convert
    - This and any other error returns query inside "unrecognized" to client
- Input parameter errors
  - This returns an error message to client


src/server/enrichment/validation/index.js
```diff
@@ -5,100 +5,144 @@ const { validTargetDb } = require('./validity-info');
...
-};
+const GCONVERT_URL = 'https://biit.cs.ut.ee/gprofiler_archive3/r1741_e90_eg37/web/gconvert.cgi';
+const FETCH_TIMEOUT = 5000; //ms
 
-class InvalidInfoError extends Error {
-  constructor(invalidOrganism, invalidTargetDb, message) {
-    super(message);
-    this.invalidTargetDb = invalidTargetDb;
-    this.invalidOrganism = invalidOrganism;
+class InvalidParameterError extends Error {
```
d2fong: We might not need an invalid parameter error here.  We could just use a normal error

src/server/enrichment/validation/index.js
```diff
@@ -5,100 +5,144 @@ const { validTargetDb } = require('./validity-info');
...
-            geneInfo.add(JSON.stringify({initialAlias: initialAlias, convertedAlias: convertedAlias}));
-          }
-        });
-        for (const initialAlias in duplicate) {
-          duplicate[initialAlias] = Array.from(duplicate[initialAlias]);
+  return form;
+};
+
+const bodyHandler = body =>  {
+  const geneInfoList = _.map(body.split('\n'), ele => { return ele.split('\t'); });
```
maxkfranz: This is better than before, but maybe we could have a student do some refactoring on this function.  It could be cleaned up and optimised a bit more

maxkfranz: Sometime later -- outside of this pr

PR #809, opened 2018-06-04
Fixed Error with tooltip Display

MichaelWrana:
Some pid pathways have entities with `null` data array.  This was causing the tooltip display to throw an error in console.  This PR fixes some errors being thrown as code assumed data array was `[]` not `null`


MichaelWrana:
Wrong Branch

PR #810, opened 2018-06-04
Fixed Error with tooltip Display

MichaelWrana:
Some pid pathways have entities with `null` data array. This was causing the tooltip display to throw an error in console.

This PR should fix some errors caused by code assuming data array was`[]` not `null`

### DIFF
--- a/src/client/common/cy/tooltips/publications.js
+++ b/src/client/common/cy/tooltips/publications.js
@@ -76,7 +76,7 @@ function getPublications(data) {
       data[i][1] = data[i][1].filter(checkForCitation);
     }
   }
-  if(data && databaseInfo !== [])
+  if(data && databaseInfo.length > 0)
     data.push([["Database IDs"],databaseInfo]);
 
   return new Promise(function (resolve, reject) {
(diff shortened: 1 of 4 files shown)


src/client/common/cy/tooltips/publications.js
```diff
@@ -76,8 +76,8 @@ function getPublications(data) {
       data[i][1] = data[i][1].filter(checkForCitation);
     }
   }
-
-  data.push([["Database IDs"],databaseInfo]);
+  if(data && databaseInfo !== [])
```
maxkfranz: `databaseInfo !== []` is always false as it's a reference/address comparison

src/server/graph-generation/biopax-metadata/index.js
```diff
@@ -74,9 +74,18 @@ function buildBioPaxTree(entity) {
   if (!(xref))
     xref = [];
```
jvwong: _.concat?

src/server/graph-generation/biopax-metadata/index.js
```diff
@@ -85,16 +94,36 @@ function buildBioPaxTree(entity) {
...
     if (!(refElement))
       continue;
+
+    //create list of xrefs for eref
+    if(keyName === "EntityReference"){
+      ErefXref = refElement['xref'];
+      if (typeof ErefXref == 'string')
+      ErefXref = [ErefXref];
+      if (!(ErefXref))
+      ErefXref = [];
```
jvwong: _.concat?

Issue #800, opened 2018-05-31
Create toolbar section for enrichment


CaitlinOCallaghan:
## Desired UI for toolbar 

![image](https://user-images.githubusercontent.com/38890251/40800312-06071cd6-64de-11e8-8222-65f950c67753.png)

## Task

Create toolbar section
- Title
- PC logo

maxkfranz:
It might be a bit confusing with two visible search bars.  Should this new search bar just be an addition to the search page rather than the results page?

CaitlinOCallaghan:
Max has a good point regarding the possible confusion. My only concern is that if the gene input is on a different page, then modifying the list could be inconvenient. We could collapse and expand the second search bar (like the live PC) since it is likely that users will only explore the tools section after already inputing their genes. But after looking at the design, I feel like it could be even more confusing. 

![image](https://user-images.githubusercontent.com/38890251/40842930-62bd62c4-657d-11e8-9643-2f3fdfd608e6.png)

Also, should the second search bar be for pathway names or for genes involved? I personally think a gene search to identify which pathways contain a specific gene would be more useful to users.

jvwong:
The first bar ('Enter one gene') is really a data input for the pathway enrichment analysis. The second is more of a 'find'.  So it likely makes sense to remove the magnifying glass icon from the former.

CaitlinOCallaghan:
Rather than a  'search' button with a magnifying glass, would we want to replace it with a 'submit' button? 

![image](https://user-images.githubusercontent.com/38890251/40854174-92b06596-659e-11e8-87a0-b2abe0309222.png)

CaitlinOCallaghan:
This is what I have so far: 

![image](https://user-images.githubusercontent.com/38890251/40858265-c47d9bd6-65ab-11e8-8ead-39a662008aca.png)

The line "|" is a result from the parent base-network-view.js file rending.

Or we can format the title instead as:

![image](https://user-images.githubusercontent.com/38890251/40860003-b74d8006-65b1-11e8-868b-d34f432f0eb2.png)

maxkfranz:
> My only concern is that if the gene input is on a different page, then modifying the list could be inconvenient. We could collapse and expand the second search bar (like the live PC) since it is likely that users will only explore the tools section after already inputing their genes.

How about hiding the entire toolbar until after the genes list has been entered?  I assume there's no graph until after the gene list is submitted, so the toolbar wouldn't be useful at first anyway.

[...]

PR #812, opened 2018-06-05
Toolbar skeleton enrich issue #800

CaitlinOCallaghan:
Skeleton for enrichment toolbar
NOTES: 

- Will only render if componentDidMount(){} is commented out - found in base-network-view.js lines 56-72

- The search bar and input bar are currently both visible, but this will be addressed once the other features of the app are developed, such as network loading


src/client/common/components/base-network-view/base-network-view.js
```diff
@@ -163,16 +164,40 @@ class BaseNetworkView extends React.Component {
...
 
+    //display pathway and database names
+    const metadataTitles = h('h4',[
+      h('span', state.networkMetadata.name),
+      ' | ',
+      h('a', state.networkMetadata.datasource)
+    ]);
+
+    //display human icon indicating human genome, gene input bar, and submit button
+    const geneInputBar =  [
```
jvwong: To me, a `geneInputBar` feels like an item that is specific to a particular view. One idea is to let views control the entire `title-container` and default to the metadataTitles?

src/client/common/components/base-network-view/base-network-view.js
```diff
@@ -178,24 +178,10 @@ class BaseNetworkView extends React.Component {
...
-    ];
-
-    // if 'useGeneInput' is true from index file, input bar will be rendered in 'div.title-container'
-    // otherwise pathway name and database will be in title bar
+    // if 'useCustomTitleContainer' is true from index file, unique title from idex's 'titleContainer' will render in 'div.title-container'
+    // default: metadata pathway name and database in title bar
     const displayInfo = [
-      (componentConfig.useGeneInput ?  geneInputBar : metadataTitles)
+      (componentConfig.useCustomTitleContainer ?  this.props.titleContainer : metadataTitles)
     ];
```
jvwong: Do we need extra boolean `useCustomTitleContainer`? Is it possible to put `titleContainer` on componentConfig and just check if it's there?

Issue #814, opened 2018-06-06
Send Input


CaitlinOCallaghan:
- [ ] Make 'submit' button functional - OnClick 'submit' sends input data

- [ ] Convert input data to readable array with all elements in the same col

- [ ] Create [`ServerAPI`](https://github.com/PathwayCommons/app-ui/blob/2b2fdda05fe3f14ea441eef2c452b825f92b0db5/src/client/services/server-api/index.js) helper method for enrichment endpoints

- [ ] Send input as 'gene [] ' to validation services

- [ ] Store result in component state

PR #820, opened 2018-06-08
Create ServerAPI helper method for enrichment endpoints #814

CaitlinOCallaghan:
Addresses portion of #814

'geneQuery' method returns 'enrich' method for validation
if/else case removed

### DIFF
--- a/src/client/services/server-api/index.js
+++ b/src/client/services/server-api/index.js
@@ -33,18 +33,14 @@ const ServerAPI = {
   },
 
   enrich(query, type){
-    if(query.genes.length>=1){
-      return fetch(`/api/${type}`, {
-        method:'POST',
-        headers: {
-          'Accept': 'application/json',
-          'Content-Type': 'application/json'
-        },
-        body:JSON.stringify(query)
-      }).then(res => res.json());
-    }else{
-      return Promise.resolve({geneInfo:[],unrecognized:[]});
-    }
+    return fetch(`/api/${type}`, {
+      method:'POST',
+      headers: {
+        'Accept': 'application/json',
+        'Content-Type': 'application/json'
+      },
+      body:JSON.stringify(query)
+    }).then(res => res.json());
   },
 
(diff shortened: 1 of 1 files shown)


src/client/services/server-api/index.js
```diff
@@ -32,19 +32,19 @@ const ServerAPI = {
     return fetch(`/pc-client/querySearch?${qs.stringify(queryClone)}`, defaultFetchOpts).then(res => res.json());
   },
 
+  enrich(query, type){
```
jvwong: `enrichmentAPI`

src/client/services/server-api/index.js
```diff
@@ -32,19 +32,19 @@ const ServerAPI = {
...
+    return fetch(`/api/${type}`, {
+      method:'POST',
+      headers: {
+        'Accept': 'application/json',
+        'Content-Type': 'application/json'
+      },
+      body:JSON.stringify(query)
+    }).then(res => res.json());
+  },
```
jvwong: Let's make a habit of putting a `catch` after these.

CaitlinOCallaghan:
Since this is for the enrichment API, can we send the error code 400 and message for the catch? Please let me know if I should change the contents of the catch.

jvwong:
It's unlikely to be a Bad Request (400). You can try [the solutions from the docs](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch#Checking_that_the_fetch_was_successful) if you want.

src/client/services/server-api/index.js
```diff
@@ -42,9 +42,7 @@ const ServerAPI = {
       body:JSON.stringify(query)
     })
     .then(res => res.json())
-    .catch((err) => {
-      res.status(400).send(err.message);
-  },
+    .catch((err) => { console.error('Error:', err); },
```
jvwong: Is this missing closing brackets?

Issue #813, opened 2018-06-06
Create working input - enrichment


CaitlinOCallaghan:
## Desired input design

![image](https://user-images.githubusercontent.com/38890251/41002088-c5da52c0-68e0-11e8-8574-438ab5ddb65f.png)

## Desired input function

- Send input
             OnClick submit --> Convert input text to proper array format --> Send gene input to 
             enrichment validation service
- Receive validation result
             If unrecognized && duplicate empty, continue to analysis
             If !empty, provide user feedback indicating invalidity ( unrecognized, duplicate )

PR #824, opened 2018-06-11
#813 Preliminary Input Error Check Enrichment

CaitlinOCallaghan:
Preliminary design that checks gene input and returns alerts.

### DIFF
--- a/src/client/features/enrichment/index.js
+++ b/src/client/features/enrichment/index.js
@@ -1,8 +1,8 @@
 const React = require('react');
 const h = require('react-hyperscript');
-//const queryString = require('query-string');
 const _ = require('lodash');
 //const Loader = require('react-loader');
+//const queryString = require('query-string');
 
 // const hideTooltips = require('../../common/cy/events/click').hideTooltips;
 // const removeStyle= require('../../common/cy/manage-style').removeStyle;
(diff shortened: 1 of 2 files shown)


src/client/features/enrichment/index.js
```diff
@@ -42,11 +95,22 @@ class Enrichment extends React.Component {
 
   geneInputSubmission(input){
     const geneArray = input.split(/\n/g);
+    if(geneArray.length < 5) return alert("Please input 5 or more tokens");
+    if(geneArray.length > 200)return alert("Please input 200 or less tokens");
     const inputObject = {genes: _.pull(geneArray,"")};
```
jvwong: Unnecessary restrictions. Remove.

src/client/features/enrichment/index.js
```diff
@@ -21,6 +21,59 @@ const enrichmentConfig={
   useSearchBar: true
 };
 
+function validateInput(query, array, object)
```
jvwong: Add comments here. Take this opportunity to list the task(s) you wish to accomplish here 

Also
- Change function name
- More informative parameter names

CaitlinOCallaghan:
The input box has been updated to provide feedback, but it still needs work. It's a bit choppy because it returns the tokens in a different order than entered, and there are currently some issues when resubmitting. I will look into fixing these problems. 

Also, do we want to differentiate the response if a gene is duplicate vs unrecognized?

CaitlinOCallaghan:
One thing to consider is that if the user puts a space after the gene they typed, ( ex: "tp53 " instead of "tp53" , the validation service will treat the token as unrecognized which can be frustrating for the user

jvwong:
> One thing to consider is that if the user puts a space after the gene they typed, ( ex: "tp53 " instead of "tp53" , the validation service will treat the token as unrecognized which can be frustrating for the user

So this argues for a little helper function that cleans input...

maxkfranz:
We should definitely do a `String.prototype.trim()` on each line.

PR #825, opened 2018-06-11
Refactor Interaction Page

q4wen:
Motivation:

To make the page faster and the code more readable.

The pull request:

- Moves graph data parsing from front end `client/features/interactions/index.js` to back end `server/graph-generation/interaction/index.js`;

- Cleans some nested functions and adds comments;

- Leaves a TODO function for caching implementation in the future.


IgorRodchenkov:
I'll test this tomorrow.. and probably merge into development and v1.0.0 branches.

src/client/features/interactions/index.js
```diff
@@ -56,171 +56,66 @@ class Interactions extends React.Component {
...
+     const state = this.state;
+     const cy = this.state.cy;
+     const categories = state.categories;
+     const filters=state.filters;
+     _.forEach(filters,(value,type)=>{
+       const edges = cy.edges().filter(`.${type}`);
+       const nodes = edges.connectedNodes();
+       edges.length?
+       categories.set(type,{edges:edges,nodes:nodes}):
+       (categories.delete(type),delete filters[type]);
```
jvwong: What does `(categories.delete(type),delete filters[type]);` accomplish?

q4wen: It is deleting unused filter. We have three filters (binding, phosphorylation and expression). If none of the nodes falls in one filter, it will be removed. I have not modified the actual "front end" code yet. I guess the function should be working.

src/client/features/interactions/index.js
```diff
@@ -56,66 +56,20 @@ class Interactions extends React.Component {
...
+    ServerAPI.getInteractionGraph({sources:sources})
+      .then(result=>{
+        const layoutConfig = getLayoutConfig('interactions');
+        const network= result.network;
+        const metaData = result.metaData;
+        this.setState({
+          componentConfig : interactionsConfig,
+          layoutConfig : layoutConfig,
+          networkJSON : network,
+          networkMetadata : metaData.networkMetadata,
```
jvwong: Let's drop `networkMetadata` and it's creation (see below `src/server/graph-generation/interaction/index.js`).  All of this was intended to populate the  'Info' panel - however, it doesn't add much value, some of the approaches are fragile (traverse), and adds unnecessary bloat.

src/server/graph-generation/interaction/index.js
```diff
@@ -0,0 +1,170 @@
...
+      return 'Phosphorylation';
+    case 'controls-expression-of':
+      return 'Expression';
+    default:
+      return '';
+  }
+}
+
+function getInteractionInfoFromPC(sources) {
+  const idArray = _.uniq(_.concat([], sources)); //convert sources to array
```
jvwong: As mentioned above, let's strip out the anything that populates `metaData.networkMetadata`.

src/server/graph-generation/interaction/index.js
```diff
@@ -0,0 +1,170 @@
...
+        metaData : metaData
+      };
+    }).catch((e)=>{
+      logger.error(e);
+      return 'ERROR : could not retrieve graph from PC';
+    });
+  });
+}
+
+function getGeneIdFromPC(source) {
```
jvwong: Strip `getGeneIdFromPC` as part of removing `metaData.networkMetadata`. This is fragile anyways (traverse and string subsettting breaks for many cases).

jvwong:
Related to my suggestions: Remove the info panel in the interactions app. I don't see much value there right now (it just fetches data form NCBI about the source).

q4wen:
> Related to my suggestions: Remove the info panel in the interactions app. I don't see much value there right now (it just fetches data from NCBI Gene about the source(s)).

There is an issue if we remove those metadata creation functions. The `graph` call only returns edges and nodes of the graph, but it does not know which node is the original search. `hgncIds` in metaData is used to identify the searched genes.

In this example, MDM2 is highlighted:
screen shot 2018-06-13 at 12 33 20 pm
<br/>
If we remove metaData, the graph will become this, with MDM2 not highlighted.
<br/>
screen shot 2018-06-13 at 12 37 00 pm

Input may be used to identify the searched node, but the input can be a symbol (MDM2) or a ID (4193). The `graph` call only return nodes with its ID. If the input is a symbol, it will fail to match the searched node. I might be wrong.

@jvwong

jvwong:
Correct me if I am wrong but: For the search app, info box, a `source` parameter used for graph call on the server derives from the `officialSymbol` variable set client-side after fetching from NCBI/UniProt. Typically, this is an HGNC symbol, extracted from GeneCards records.

Can we simply use `source` as is to highlight node(s) since these are all HGNC? What if we say that you wont' get any hits when you put in some source like `http://identifiers.org/uniprot/P04637` or `7157`?

q4wen:
In the old requirement document:

> The system should be accessible
> 
> - Indirectly from the app-ui search landing page box
> - Directly via an endpoint

What is 'directly via an endpoint'? If the page can only be accessible from the landing page, using source as gene IDs should work.

jvwong:
Let's scale back and support access from the landing page.

IgorRodchenkov:
what’s it about?

IR

src/server/graph-generation/interaction/index.js
```diff
@@ -0,0 +1,139 @@
...
+
+function getInteractionInfoFromPC(sources) {
+  const idArray = _.uniq(_.concat([], sources)); //convert sources to array
+
+  const geneIds = idArray.map(source =>
+    source.includes('pathwaycommons') ? getGeneIdFromPC(source) : source.replace(/\//g,' ')
+  );
+   const metaData = {
+     networkMetadata: {
+       name : geneIds.join(','),
```
jvwong: Like before, let's put something like `Interactions - X, Y, ...`.

src/server/graph-generation/interaction/index.js
```diff
@@ -0,0 +1,139 @@
...
+      return 'Expression';
+    default:
+      return '';
+  }
+}
+
+function getInteractionInfoFromPC(sources) {
+  const idArray = _.uniq(_.concat([], sources)); //convert sources to array
+
+  const geneIds = idArray.map(source =>
```
jvwong: We need to drop this altogether and just use `source` below. As it stands, this doesn't meet its goals/fragile. It seems to be trying to accept any URI as source map it to a gene ID...

src/server/graph-generation/interaction/index.js
jvwong: `geneIds` could be replaced with `source` yes?

src/server/graph-generation/interaction/index.js
```diff
@@ -0,0 +1,139 @@
...
+       network : network,
+       metaData : metaData
+     };
+   }).catch((e)=>{
+     logger.error(e);
+     return 'ERROR : could not retrieve graph from PC';
+   });
+}
+
+function getGeneIdFromPC(source) {
```
jvwong: This can go. It's just too fragile.

q4wen:
@IgorRodchenkov 
> Instead of using fetch and building PC queries every time with config.PC_URL + command + ..., I'd use the internal pathway-commons module and its universal 'query' method. Lately, we polished and organised pc-server related code which used to spread here and there.. and now this repeats almost the same code in various places over again.

As @jvwong suggests, the PC fetch functions have been removed.

IgorRodchenkov:
Ok (perhaps I lost the track of these changes..)
I will look, again after merging this PR, and open a new issue if needed. 
Please, test this all works as expected...

Issue #834, opened 2018-06-15
Fix lint errors before v1.0.0 release


IgorRodchenkov:
(development) npm run lint:

`✖ 33 problems (33 errors, 0 warnings)`

refs #811

d2fong:
That’s a lot of problems

maxkfranz:
OK, maybe we should just worry about that for a later release then.  The students could clean up the lint errors and we could do a 1.0.1 release afterwards.

q4wen:
`console.log()` causes `Unexpected console statement  no-console` error. Should we remove `console.log(err)` or disable `no-console` rule in lint?

maxkfranz:
You shouldn't be using `console.log()` for logging errors.  There's a logger module that can be used to log to the console and to log files using winston.  It has the same interface as `console`.

There are three log levels:

- `console.log()` -- debug level, e.g. query hit an edgecase
- `console.info()` -- info level, e.g. served this query
- `console.error()` -- error level, e.g. query failed

PR #836, opened 2018-06-19
Fixed Some Lint errors

MichaelWrana:
(development) `npm run lint` : `✖ 6 problems (6 errors, 0 warnings)`

I tried to focus on the straightforward stuff like missing semicolons and unused variables.  The remaining errors might require more attention.

With these changes, `npm run test` gives `45 passing (59s)`

Refs #834

### DIFF
--- a/src/client/common/cy/layout/index.js
+++ b/src/client/common/cy/layout/index.js
@@ -1,4 +1,5 @@
-//Unused For Now
+//This function produces better results for most layouts that do not have compound support
+//unused currently.
 /*
 const pcGraphLayoutHack = (cy, layoutOpts) => {
   const complexes = cy.nodes().filter(node => node.isParent() && node.data('class') === 'complex');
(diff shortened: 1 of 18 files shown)


src/client/common/cy/layout/index.js
```diff
@@ -1,3 +1,5 @@
+//Unused For Now
```
d2fong: Please add a comment saying this is produces better results for most layouts that do not have compound support.

src/client/common/cy/tooltips/format-content.js
```diff
@@ -329,7 +329,7 @@ function generateDBLink(dbName, dbId, isDbVisible) {
  *    </div>
  * </div>
  */
-function noDataWarning(name) {
+function noDataWarning() {
```
d2fong: Please add a TODO comment here saying this function needs to be reworked.

src/server/routes/rest.js
```diff
@@ -22,6 +21,7 @@ router.post('/submit-layout', function (req, res) {
         res.json(package);
       });
   } else {
+    let errorMsg = "Issue in exposing REST endpoint for controller.submitLayout.  See /src/server/routes/rest.js";
```
d2fong: Can you revert all the changes in this file?  I handled these in my branch.

PR #837, opened 2018-06-19
Clean up lint error

q4wen:
Ref: #834 
The clean-up includes:

- Remove unused variables and modules

- Remove useless escape character in regex

- Replace `console.log()  ` with winston logger

- Replace upper case colour code with lower case

- Remove variable unit with zero value in css files

- Format css files (add new line to the end of files, spacing before and after comments)


q4wen:
#836 Repeating PR

Issue #815, opened 2018-06-06
Receive and Interpret Validation Response


CaitlinOCallaghan:
If ' unrecognized [] ' && ' duplicate [] ' empty, continue to analysis ( do nothing for now ) 
If arrays !empty, provide user feedback indicating invalidity (unrecognized and/or duplicates)

- [ ] underline tokens violating input conventions
- [ ] pop up message indicating type of invalidity 
                  if multiple instances of same type of infraction, only provide one message
                  if input token count < 4, ask for larger input 

## Test cases: 

1. All valid
   genes: [ 
      "TP53",
      "ATM",
      "MCM3",
      "XPO1"
   ]

2. One unrecognized
   genes: [ 
      "notAGene",
      "TP53",
      "ATM",
      "MCM3",
   ]

3. One set duplicate
   genes: [ 
      "TP53",
      "TP53",
      "ATM",
      "MCM3",
   ]

4. One set duplicate and one unrecongnized
   genes: [      
      "notAGene",
      "TP53",
      "TP53",
      "ATM",
   ]

5. All valid, but not enough tokens
   genes: [ 
      "TP53",
      "ATM",
      "MCM3",
   ]

jvwong:
Do we mark valid IDs explicitly or just leave as is.

CaitlinOCallaghan:
For now, leave as is and once everything is up and running we can add that feature

jvwong:
We're all realizing how non-trivial this task is. Let's reduce the scope of this component for now:

- [x] Accept text input
- [x] Display unrecognized tokens
- [ ] Provide an array of unique recognized gene identifiers (i.e. for parent component)

PR #839, opened 2018-06-20
Functional Invalid Token Checking Enrichment #813 , #815

CaitlinOCallaghan:
Addresses #813, and is a MVP for #815. 

- Token input methods are located in file token-input.js 
- Tokens are submitted through a div --> this will be iterated upon and improved in the future (use of an external library like slate)
- Tokens are stored on state in the form of a map where keys are token names and values (true or false) correspond to token validity 
- Input alterations are allowed and only new tokens are sent for validation
- Invalid tokens are currently displayed in a div that updates as invalid tokens are removed from the list --> the mechanism for user feedback will be iterated upon and improved in the future, tied to input box changes

NOTE: code will only run if ComponentDidMount() is commented out in base-network-view.js since the network has not been set up yet


maxkfranz:
> Tokens are submitted through a div

Is there a textarea in the div, or are you using contenteditable?  It might be simpler to use a textarea at this point.

CaitlinOCallaghan:
I'm currently using contenteditable, but I can look to change it to a textarea (I agree, textarea will probably be simpler).

src/client/features/enrichment/token-input.js
```diff
@@ -0,0 +1,75 @@
+//const React = require('react');
```
jvwong: I think if you turned this into it's own React component, complete with all of its own display and logic, this would be much more logical/readable.

jvwong: To clarify, I was referring to an 'gene input' component.

maxkfranz: Yes, this should be using a React component with a `render()` function.

src/client/features/enrichment/token-input.js
```diff
@@ -0,0 +1,75 @@
...
+
+  constructor(state) {
+    this.state = state;
+  }
+
+  //called onClick 'submit'
+  //seperate div text input into an array, remove all elements that have already been validated
+  //this method will be altered later depending on the type/format of the final input box (slate or some other library)
+  parseTokenList() {
+    this.state.query = document.getElementById('gene-input-box').innerText;
```
maxkfranz: Shouldn't be reading from the DOM.  Use a listener in render()

src/client/features/enrichment/token-input.js
```diff
@@ -0,0 +1,75 @@
...
+    this.state.tokenData.forEach( (value, key) => {
+      if (tokenList.includes(key)) tokenList = _.pull(tokenList, key);
+    });
+    //send only new tokens to validation
+    this.retrieveValidationAPIResult(tokenList);
+  }
+
+  //call validation service API to retrieve unrecognized tokens as an array
+  retrieveValidationAPIResult(tokensToValidate){
+    ServerAPI.enrichmentAPI({genes: _.pull(tokensToValidate,"")}, "validation").then((result) => {
```
maxkfranz: Promises are not returned and so can't be chained

src/client/features/enrichment/token-input.js
```diff
@@ -67,9 +70,44 @@ class TokenInput {
...
+ render() {
+
+  return (
+    //titleContainer: [
+      [h('h4', [
+        h('span', 'Pathway Enrichment   '),]),
+      h('img', {
+        src: '/img/humanIcon.png'
+        }),
+      h('div.gene-input-container', [
```
jvwong: Everything above the `gene-input-container` should go in parent. Everything below can stay.

src/client/features/enrichment/token-input.js
CaitlinOCallaghan: I think the ServerAPI.enrichmentAPI calls upon the validation index.js which then returns a promise

src/client/features/enrichment/index.js
```diff
@@ -30,22 +32,22 @@ class Enrichment extends React.Component {
         datasource: '',
         comments: []
       },
-      query: '',
-      titleContainer: []
+      titleContainer: [],
+      invalidTokenContainer: [],
```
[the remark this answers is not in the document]
CaitlinOCallaghan: It was initially used to hide the toolbar and replace that section with a feedback box. In the latest commit, I changed the structure to be more logical and to make it easier to open and close both the toolbar and feedback box.

CaitlinOCallaghan:
Currently, the input component is in a separate file than the main index.js . I have an alternative version saved on my desktop that has everything in the index file, if that would be preferable, but it's about 175 lines long. In both versions, there aren't any console warnings or errors.

src/client/features/enrichment/token-input.js
```diff
@@ -0,0 +1,108 @@
...
+    let tokenList = this.state.query.split(/\s/g);
+    this.state.tokenData.forEach( (value, key) => {
+      if (tokenList.includes(key)) tokenList = _.pull(tokenList, key);
+    });
+    this.retrieveValidationAPIResult(tokenList);
+  }
+
+  //call validation service API to retrieve validation result in the form of []
+  //****** NOTE: currently only checking for unrecognized tokens, not duplicates
+  retrieveValidationAPIResult(tokensToValidate){
```
jvwong: Suggestion: Keep this function, and remove all the others.

src/client/features/enrichment/token-input.js
```diff
@@ -0,0 +1,80 @@
...
+  //store 'gene-input-box' contents on state
+  handleChange(e) {
+    this.state.query = e.target.value;
+  }
+
+  //call validation service API to retrieve validation result in the form of []
+  retrieveValidationAPIResult(){
+    //reset state values
+    this.state.unrecognizedTokens = [];
+    this.state.validTokens = [];
```
jvwong: do not modify state directly (https://reactjs.org/docs/state-and-lifecycle.html#do-not-modify-state-directly)

src/client/features/enrichment/token-input.js
```diff
@@ -0,0 +1,80 @@
...
+    //send all tokens to validationAPI
+    ServerAPI.enrichmentAPI({genes: _.pull(tokenList,"")}, "validation").then((result) => {
+      this.interpretValidationResult(tokenList, result.unrecognized);
+      });
+  }
+
+  //store validation data on state as arrays [validTokens] and [unrecognizedTokens]
+  //lift [validTokens] to parent file index.js
+  interpretValidationResult(tokenList, unrecognizedTokens){
+    tokenList.forEach((element) => {
```
jvwong: Is all of this necessary? 
You already have 1. unrecognized 2. recognized tokens. Also I believe you're attempting to set state on each iteration of the loop which is probably slow.

Issue #830, opened 2018-06-15
Add hide-by-degree slider to interactions page


MichaelWrana:
## Description of new feature

- Add a new slider to the `interaction` view
- When slid, nodes would disappear and reappear based on their degree

## Motivation for new feature

- Help users clear unwanted nodes from `interaction` view
- Improve load time by starting the slider with some non-zero value
- Consider a page like this: http://appsbeta.pathwaycommons.org/interactions?source=TP53

MichaelWrana:
This is how i've designed the slider:

![screen shot 2018-06-19 at 5 00 27 pm](https://user-images.githubusercontent.com/25777239/41623968-5cb8bc5c-73e2-11e8-8441-d920dd90b66c.png)


Any suggestions for improvement? Or does it look good as-is

jvwong:
@maxkfranz just mentioned to me there is styling and code elements for sliders in app-ui/factoid...

maxkfranz:
Yes, you can use the built-in HTML input tag and it will be styled for you by the existing stylesheet to be consistent cross-browser/cross-OS: 

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/range

PR #840, opened 2018-06-21
Degree-Based Slider

MichaelWrana:
This PR adds a slider to the `interaction` view which hides nodes solely based on degree.

Refs #830 

Appearance:
image


MichaelWrana:
Closing, new slider created using centrality instead of degree: #853

Issue #833, opened 2018-06-15
Add hide/show nodes feature to view page


MichaelWrana:
## Description of new feature
- Allow one to select a subset of components of a pathway 
    - And view those selected components
    - And view those unselected components

## Motivation for new feature

This will help users clear clutter from the `view` page by giving them the ability to select and hide nodes.

MichaelWrana:
Replaced with network filtering and slider

PR #841, opened 2018-06-21
Added hide/show node buttons to Network View

MichaelWrana:
Refs #833 
Current Implementation:

1. User presses shift+left-click+drag to select an area on screen, and nodes within that area are highlighted:

image

2. User can then press the struck-through eye icon to hide those selected nodes:

image
image

3. Steps 1 & 2 can be repeated until no nodes remain in the graph.

4. Pressing the 'eye' icon will re-show any nodes that have been hidden previously, reproducing the original graph.


MichaelWrana:
Need to fix edges not reappearing issue, don't pull for now.

jvwong:
Offline discussion with @jvwong and @MichaelWrana: This feature brings to forefront several bugs that impact function. We will list, triage in separate issues.

src/client/common/components/base-network-view/config/index.js
```diff
@@ -21,6 +21,35 @@ const resetToDefaultLayout = (props) => {
...
 
+//This list of nodes is created from shift+drag box select.  See box-select.js
+const hideSelectedNodes = (props) => {
+  const cy = props.cy;
+  let nodesToHide = cy.selectedNodesToHide;
+  for(let i in nodesToHide){
+    let node = nodesToHide[i];
+    node.hide();
+  }
+  cy.selectedNodesToHide = [];
```
maxkfranz: This is the relevant line.  Shouldn't be setting things in `cy`...

src/client/common/components/base-network-view/config/index.js
```diff
@@ -21,6 +21,35 @@ const resetToDefaultLayout = (props) => {
   cy.layout(props.layoutConfig.defaultLayout.options).run();
 };
 
+//This list of nodes is created from shift+drag box select.  See box-select.js
+const hideSelectedNodes = (props) => {
+  const cy = props.cy;
+  let nodesToHide = cy.selectedNodesToHide;
+  for(let i in nodesToHide){
+    let node = nodesToHide[i];
+    node.hide();
```
maxkfranz: Use classes.  Do not use `.show()` or `.hide()`

maxkfranz:
show() and hide() won't allow for composition with other show/hide features like the expand/collapse button.  It makes things fragile.  It's better to use classes so your stylesheet can handle the overlapping cases in one place

MichaelWrana:
This won't work with the new pathway viewer changes

PR #848, opened 2018-06-22
swapped formData for background

jvwong:
Fixing missing parameter for  #847.


jvwong:
OK closing in favour of #851

Issue #864, opened 2018-07-03
Correctly spell database names in display text   


jvwong:
Cases:

- Landing box
  - [x] `UniProt`
  
![image](https://user-images.githubusercontent.com/4706307/42227369-5cff6274-7eaf-11e8-98a4-fd66fae63df9.png)

- Pathway view tooltips
  - [ ] `UniProt`
  - [ ] `HGNC` rather than `HGNC Symbol`
![image](https://user-images.githubusercontent.com/4706307/42227438-86fa3d2e-7eaf-11e8-8914-d7455689544e.png)

MichaelWrana:
There are separate cases for links to be named as "HGNC" vs "HGNC Symbol" : 

![screen shot 2018-07-04 at 10 15 46 am](https://user-images.githubusercontent.com/25777239/42282033-3e47f7fa-7f73-11e8-91a8-195456e0483f.png)


Should I combine or leave as-is?

jvwong:
Let's leave this alone. Its not worth changing.

jvwong:
The tooltip link to UniProt is still spelled incorrectly:

![image](https://user-images.githubusercontent.com/4706307/43078028-30518ec0-8e57-11e8-848c-3ab7ff5801e1.png)

Issue #856, opened 2018-06-28
Gene Cards gene linkout returning a 404  


jvwong:
## Environment info

- Browser version (client) : n/a
- Node.js version (server) : 8

## Current (buggy) behaviour

Link to Gene Cards for a gene returns a 404 page. 

## Desired behaviour

Return the Gene cards record.

## Minimum steps to reproduce

Click infobox link to Gene Cards for search of [uniprot:Q99988](http://appsbeta.pathwaycommons.org/search?gt=0&lt=250&q=uniprot%3AQ99988&type=Pathway)

see https://www.ebi.ac.uk/miriam/main/collections/MIR:00000323

PR #869, opened 2018-07-03
Update landing page links

q4wen:
Ref: #856  #864
- Gene Card link is removed

- Uniprot is changed to UniProt

screen shot 2018-07-03 at 12 12 33 pm

### DIFF
--- a/src/client/features/search/landing-box.js
+++ b/src/client/features/search/landing-box.js
@@ -9,10 +9,9 @@ const _ = require('lodash');
 
 //usedDatabases=[['Uniprot lookup name',{configName:,gProfiler:}]]
 const usedDatabases=new Map ([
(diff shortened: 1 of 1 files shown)


jvwong:
1. This inadvertently removed the 'Official Symbol' field. This is part of #875.

2. We should use Gene Cards recommended link instead:

> Linking to particular GeneCards
>If you are a web developer and you want to make links to particular GeneCards, use 
"https://www.genecards.org/cgi-bin/carddisp.pl?gene=XXX" (with XXX being the gene symbol).

>  To link using an external id, use "https://www.genecards.org/cgi-bin/carddisp.pl?id=YYY" 
(where YYY is a UniProt accession, EntrezGene id, Ensembl id, or HGNC id).
> Optionally, a source for the id can be specified by adding "&id_type=" to the url. The possible types are "uniprot", "entrezgene", "ensembl", and "hgnc". To avoid confusion, it is strongly recommended to use the id_type parameter, especially for numeric ids.

Issue #845, opened 2018-06-21
Include number of genes in gene set / nodes - Enrichment 


CaitlinOCallaghan:
- Amend enrichment visualization API service to provide number of genes

jvwong:
Node size should be proportional to number of genes.

jvwong:
This should include corresponding changes to Swagger

PR #880, opened 2018-07-05
Visualization service update

q4wen:
Ref: #845 

Each node in the graph in visualization service includes number of genes. Swagger is updated correspondingly.

screen shot 2018-07-05 at 3 08 14 pm


jvwong:
I vote 'gene-count'. @CaitlinOCallaghan?

maxkfranz:
There should be no dashes in any key names.  Use camelCase not dash-case.  @q4wen and @CaitlinOCallaghan, would one of you create a ticket for this please?

Issue #870, opened 2018-07-04
Hover + zoom leads to 'supersized' glyphs in viewer


jvwong:
Consider a pathway that is zoomed out: Highlighting a glyph (by mousing over) then zooming in leads to glyphs that are very large. 

One would expect the glyphs to be the same size at any time during the zoom.

I have reproduced an example for the [Transcriptional activation of cell cycle inhibitor p21 | Reactome](http://localhost:3000/view?uri=http%3A%2F%2Fidentifiers.org%2Freactome%2FR-HSA-69895) by zooming out, hovering over process node, then zooming in while maintaining hover: 

![zoombug](https://user-images.githubusercontent.com/4706307/42249953-04df751e-7efc-11e8-98cd-2d80a628f8ad.gif)

maxkfranz:
We should probably just rework the highlight this and nix the font changes altogether: 

Hover highlight should use opacity #868

Issue #868, opened 2018-07-03
Hover highlight should use opacity


maxkfranz:
## Description of new feature

The hover highlight should not change node or edge colour.  It should instead fade out the other edges and nodes.

## Motivation for new feature

You can still see the edge colours.

jvwong:
I don't see this functioning on the SBGN/pathway views but it works on ball-stick views.

MichaelWrana:
It looks like @d2fong is doing a major rework of the pathway view, which I think is related to the hover stuff.  Maybe it no longer works because pathway view now has its own event files?

MichaelWrana:
Update: Should work for SBGN view now with #925

jvwong:
I find that the highlight on hover for edges rather cumbersome, especially when there are a lot of elements on the screen. Would it be better to highlight on node only?

maxkfranz:
I agree nodes only is probably less jittery.  It's too easy to inadvertently hover over an edge in dense graphs.

PR #889, opened 2018-07-11
Modified Hover to only change opacity

MichaelWrana:
Refs #868 #870 
Reworked the hover functionality to only modify opacity.
Some screenshots of how the new hover will look:
screen shot 2018-07-11 at 11 33 09 am
screen shot 2018-07-11 at 11 32 45 am
screen shot 2018-07-11 at 11 32 30 am


d2fong:
Neat

d2fong:
Waiting on Jeff/Gary feedback

maxkfranz:
Re. the code, it shouldn't be applying bypasses.  It should be using the stylesheet.  This would drastically simplify the code.

Classes:

- highlighted
- unhighlighted

Adding highlight:

```js
elesToHighlight.addClass('highlighted');

cy.elements().difference(elesToHighlight).addClass('unhighlighted');
```

Clearing:

```js
cy.elements().removeClass('highlighted unhighlighted');
```

MichaelWrana:
When I want to create a new CSS class, e.g. `highlighted` and `unhighlighted` and add it to different network views, it looks like I have to add them all identically to:

`src/client/common/cy/stylesheet.js`
`src/client/common/cy/interactions-stylesheet.js`
`src/client/common/cy/enrichment-stylesheet.js`

This works, but seems pretty prone to breaking in future (eg someone creating a new view and not adding these classes).  Maybe we should create a generic `stylesheet.js` somewhere that applies to all views?  Or this already exists and I missed it?

maxkfranz:
There should be a default stylesheet, and the other specific stylesheets can extend them.  If you use a JSON stylesheet, then

```
let stylesheet = [ /* basic style that applies to all views */ ];

let specificStylesheet = stylesheet.concat([ /* new style blocks on top... */ ])
```

jvwong:
This is a great solution, leaves the door open to better use highlight for other tasks (e.g. selection). I can't run branch locally b/c of conflicts...

jvwong:
@MichaelWrana I don't find this feature present after the merge?

Issue #902, opened 2018-07-13
correlate node size with 'geneCount'


CaitlinOCallaghan:
## Desired behaviour

- correlate node size with 'geneCount'
- limit smallest node size to 30

CaitlinOCallaghan:
Is there a mathematical relationship that we want to use for relating the geneCount to node size?

Currently I have it set to the square root function  " 5* sqrt( x - 1 ) + 30 " , https://www.desmos.com/calculator/w6ux3lhwjg , where x is geneCount and if geneCount > 1000, then x is set to 1000.

PR #903, opened 2018-07-16
 geneCount mapped to node size #902

CaitlinOCallaghan:
- mathematically map geneCount to enrichment node size
- default and smallest node size = 30
- largest node size = 188 ( when geneCount >= 1000 )
- increase nodeRepulsion so nodes do not overlap

### DIFF
--- a/src/client/common/cy/enrichment-stylesheet.js
+++ b/src/client/common/cy/enrichment-stylesheet.js
@@ -28,7 +28,7 @@ const enrichmentStylesheet=cytoscape.stylesheet()
 .css({
   'font-size': 20,
   'color': 'black',
-  'background-color': '	#00bfff', //TODO: Colored accoriding to p-value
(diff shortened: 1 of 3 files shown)


src/client/common/cy/enrichment-stylesheet.js
```diff
@@ -1,6 +1,15 @@
 const cytoscape = require('cytoscape');
 
-const iStylesheet=cytoscape.stylesheet()
+function getNodeSize( geneCount ){
+  let size = ( geneCount <= 1000 ) ? mapGeneCountToSize(geneCount) : mapGeneCountToSize(1000);
```
d2fong: It would be easier to understand this code if you did something like:
```js
return mapGeneCountToSize(Math.min( geneCount, 1000 ));
```

CaitlinOCallaghan: That is a lot better, thank you!

Issue #857, opened 2018-06-28
Custom approach for direct fetch from genenames.org for query hgnc:<id> pattern


jvwong:
//TODO 
// https://rest.genenames.org/fetch/symbol/TP53

q4wen:
The return JSON does not contains description or other database links. If the validator fails, the landing page will only show official name, alias and a link to HGNC. Should the landing box still be displayed?

q4wen:
As describe above, if the validator fails, the landing page from HGNC will be like this:

![screen shot 2018-07-03 at 4 35 26 pm](https://user-images.githubusercontent.com/16858225/42243395-2f0bdc00-7edf-11e8-81cd-2e10cbc23742.png)

jvwong:
There **may** be other db links for UniProt, NCBI Gene (Entrez). Not always the case. 

Below for [TP53](http://rest.genenames.org/fetch/symbol/TP53) there are IDs for both:

```json
{
   ...
    "response": {
        ...
        "docs": [
            {
                "hgnc_id": "HGNC:11998",
                "symbol": "TP53",
                "name": "tumor protein p53",               
               ...
                "entrez_id": "7157",
                ...
                "uniprot_ids": [
                    "P04637"
                ],
               ...
            }
        ]
    }
}
```

The gene for a microRNA [MIRLET7A1](http://rest.genenames.org/fetch/symbol/MIRLET7A1) is a gene that doesn't have a protein (so no UniProt) but has a NCBI Gene entry (Entrez)

```js
{
   ...
    "response": {
       ...
        "docs": [
            {
                "hgnc_id": "HGNC:31476",
                "symbol": "MIRLET7A1",
                "name": "microRNA let-7a-1",
              ...
                "alias_symbol": [
                    "hsa-let-7a-1"
                ],
               ...
                "entrez_id": "406881",
             ...
        
            }
        ]
    }
}
```

Issue #855, opened 2018-06-27
Landing page: NCBI infobox is tightly coupled to validator  


jvwong:
The landing page info box for `q=ncbi:<id>` fails when validator does. 

The NCBI Gene case uses a different strategy than UniProt, which does operate even if the validator fails. 

Refer to [src/client/features/search/landing-box.js](https://github.com/PathwayCommons/app-ui/blob/development/src/client/features/search/landing-box.js#L66):

  - `getNcbiInfo` is sensitive to the `genes` parameter. The `officialSymbol` value appears to me  to  be available in ESummary's `geneResults[gene].name`.
  - consequently, probably a way around the `links`

```js
...
const getNcbiInfo = (ids,genes) => {
  const ncbiIds=_.map(ids,(search,id)=>id);
  return ServerAPI.getGeneInformation(ncbiIds).then(result=>{
    const geneResults=result.result;
    return geneResults.uids.map(gene=>{
      const originalSearch = ids[gene];
      const links=idToLinkConverter(genes[originalSearch]);
      return {
        databaseID:gene,
        name:geneResults[gene].nomenclaturename,
        function: geneResults[gene].summary,
        officialSymbol:genes[originalSearch]['Gene Cards'],
        otherNames: geneResults[gene].otheraliases ? geneResults[gene].otheraliases:'',
        showMore:{full:!(geneResults.uids.length>1),function:false,synonyms:false},
        links:links
      };
    });
  });
};
```

Other related:

- [`labeledId`](https://github.com/PathwayCommons/app-ui/blob/development/src/client/features/search/landing-box.js#L143) is declared, initialized  but never used.

jvwong:
```js
const getNcbiInfo = (ids) => {
  const ncbiIds=_.map(ids,(search,id)=>id);
  return ServerAPI.getGeneInformation(ncbiIds).then(result=>{
    const geneResults=result.result;
    return geneResults.uids.map(gene=>{
      // const originalSearch = ids[gene];
      let links = idToLinkConverter({
        "NCBI Gene": gene,
        "HGNC Symbol": geneResults[gene].name,
        "Gene Cards": geneResults[gene].name
      });
      // }
      return {
        databaseID:gene,
        name:geneResults[gene].nomenclaturename,
        function: geneResults[gene].summary,
        officialSymbol:geneResults[gene].name,
        otherNames: geneResults[gene].otheraliases ? geneResults[gene].otheraliases:'',
[...]

PR #907, opened 2018-07-17
Entity Info Box: direct fetch from data provider for query <database>:<id>

q4wen:
Refs: #855 and #857 

When the query has format <database>:<id>, entity info box fetch data from provider directly without fetching data from gProfiler.


src/client/features/search/query-entity-info.js
```diff
@@ -92,14 +71,15 @@ const getNcbiInfo = ( ids, genes ) => {
     let geneResults = result.result;
 
     return geneResults.uids.map( gene => {
-      let originalSearch = ids[ gene ];
-      let links = idToLinkConverter( genes[ originalSearch ] );
+      const originalSearch = ids[ gene ];
+      const geneId = genes ? genes[ originalSearch ] : {"NCBI Gene" : originalSearch};
```
d2fong: use single quotes for strings

src/client/features/search/query-entity-info.js
```diff
@@ -154,6 +134,52 @@ const getUniprotInfo = ids => {
...
+
+  return ServerAPI.getHgncInformation(hgncSymbol).then(result=>{
+    const geneResults = result.response.docs;
+    return geneResults.map(gene => {
+
+      let databaseId = {};
+      Object.entries(dbInfos).forEach(([k,v]) => {
+        if (gene[v.hgncName]) databaseId[v.name] = gene[v.hgncName];
+      });
+      const links= idToLinkConverter(databaseId);
```
d2fong: use consistent whitespace 1 space on each side of the =

src/client/features/search/query-entity-info.js
```diff
@@ -154,6 +134,52 @@ const getUniprotInfo = ids => {
...
+
+      return {
+        databaseID:gene,
+        name: gene.name,
+        function: '',
+        officialSymbol: gene.symbol,
+        otherNames: gene.alias_symbol.join(', '),
+        showMore:{full:!(geneResults.length>1),function:false,synonyms:false},
```
d2fong: 1 space after colons, brackets, >, etc.

```js
showMore: { full: !(geneResults.length > 1), function: false, synonyms: false }
```

src/client/features/search/query-entity-info.js
```diff
@@ -188,66 +216,60 @@ const queryEntityInfo = query => {
       uniprotIds[ entityId ] = entityId;
     } else if( isNcbiId( token ) ){
       ncbiIds[ entityId ] = entityId;
-      genes.push( entityId );
+    } else if (isHgncId( token ) ){
+      hgncId[ entityId ] = entityId;
     } else {
+      prefix = false;
```
d2fong: what is the purpose of this prefix variable?

q4wen: To determine whether the query has <database>:<id> format. If not, validate the id in gProfiler.

d2fong: So queries like:

`ncbi:1 uniprot:p06777 hgnc:tp53` will get the provider information for each of these tokens but

`ncbi:1 uniprot:p06777 hgnc:tp53 blah` will send all of these ids to the validator?  

I think it makes sense to create two lists of tokens.  One list contains ['uniprot:p06777', ... 'ncbi:1'], and the other contains ['blah'].  Get the provider specific info for each token in the first list, and validate each token from gProfiler in the second list.  @jvwong is that what you meant?

q4wen: Do we consider the case `ncbi:1 uniprot:p06777 hgnc:tp53 blah`? I assume the system only accepts either:
1. `<db>:<id>` (only one token)
2. `<ids>` (one or more tokens)

jvwong: The prefix is only valid when a single token is available. Else send to validator.

src/client/features/search/query-entity-info.js
```diff
@@ -154,6 +134,56 @@ const getUniprotInfo = ids => {
   });
 };
 
+const getHgncInfo = hgncSymbols => {
+  const hgncSymbol = Object.keys(hgncSymbols)[0];
+
+  return ServerAPI.getHgncInformation( hgncSymbol ).then( result => {
+    const geneResults = result.response.docs;
+    return geneResults.map(gene => {
```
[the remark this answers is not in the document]
q4wen: https://www.genenames.org/help/rest-web-service-help
>Fetch will not accept wildcards or multiple queries 

HGNC fetch can only accept one token and return one result, so response.dcos is always = 1. Plus `getHgncInformation()` is only called when the token is prefixed (implying there is only one token).

Maybe for consistence reason, I should wrap the `fetch` within a loop so that it accept multiple tokens like `getUniprotInfo()` and `getHgncInfo()`.

src/client/features/search/query-entity-info.js
d2fong: If that is the case, I think the code should reflect that instead of checking each token to see if it is special.

src/client/features/search/query-entity-info.js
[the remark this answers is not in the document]
maxkfranz: Promises like `fetch()` should not go in a loop.  Use `Promise.all( someList.map(doSomeFetch) )` or similar

src/client/features/search/query-entity-info.js
```diff
@@ -160,6 +139,56 @@ const getUniprotInfo = ids => {
...
 };
 
+const getHgncInfo = hgncSymbols => {
+  return Promise.all(Object.keys(hgncSymbols).map( hgncSymbol => {
+    return ServerAPI.getHgncInformation( hgncSymbol ).then( result => {
+      const geneResults = result.response.docs;
+      return geneResults.map(gene => {
+
+        let databaseId = {};
+        Object.entries(dbInfos).forEach( ( [k,v] ) => {
```
jvwong: We're missing UniProt and Gene Cards here, both of which are available in the response from `rest.genenames.org/fetch/symbol` as `uniprot_ids`, `symbol`.

src/client/features/search/query-entity-info.js
```diff
@@ -98,15 +76,16 @@ const getNcbiInfo = ( ids, genes ) => {
     let geneResults = result.result;
 
     return geneResults.uids.map( gene => {
-      let originalSearch = ids[ gene ];
-      let links = idToLinkConverter( genes[ originalSearch ] );
+      const originalSearch = ids[ gene ];
+      const geneId = genes ? genes[ originalSearch ] : { 'NCBI Gene': originalSearch };
```
jvwong: Surely links to HGNC and Gene Cards can be generated from the `name` value in the response. They `genes` parameter only seems to possibly supply a uniprot accession.

q4wen:
I guess fetchError on uniprot:XXXXX is related to #922 
Uniprot changed the default return type from JSON to XML.

src/client/features/search/query-entity-info.js
q4wen: Is HGNC symbol always the same as GeneCards id?

jvwong: https://www.genecards.org/Guide/AboutGeneCards

> Linking to particular GeneCards
> ...
> To link using an external id, use "https://www.genecards.org/cgi-bin/carddisp.pl?id=YYY" 
(where YYY is a UniProt accession, EntrezGene id, Ensembl id, or HGNC id).
> Optionally, a source for the id can be specified by adding "&id_type=" to the url. The possible types are "uniprot", "entrezgene", "ensembl", and "hgnc". To avoid confusion, it is strongly recommended to use the id_type parameter, especially for numeric ids.

jvwong: It seems that the 'gene symbol' is the hgnc approved symbol in the cases that I tried...

jvwong: [Wikipedia entry](https://en.wikipedia.org/wiki/GeneCards)
> The core gene list is based on approved gene symbols published by the HUGO Gene Nomenclature Committee (HGNC).[7][8]

src/client/features/search/query-entity-info.js
```diff
@@ -144,6 +139,7 @@ const getUniprotInfo = ids => {
```
jvwong: Consider `let matchedDb = dbInfos[ db.type ];` - will this variable ever be used (non null)?

q4wen:
Uniprot server was down this morning. I try to add error handling to fetch functions.

src/client/features/search/query-entity-info.js
q4wen: It is repeat code.

jvwong:
The uniprot issue is directly relevant to #924.

PR #915, opened 2018-07-18
Filter enrichment nodes by p-value with slider

CaitlinOCallaghan:
Implement slider to filter visible nodes by p-value
Slider default value set to 0.025 when network is rendered


src/client/common/components/base-network-view/base-network-view.js
```diff
@@ -70,10 +72,10 @@ class BaseNetworkView extends React.Component {
 
       const layout = cy.layout(initialLayoutOpts);
 
-      //set default view for filtering nodes by p_value when new network is rendered
       this.state.cySrv.loadPromise().then(() => {
-        //by default, only display nodes with p_value <= 0.025
-        this.displayDefaultNodes(0.025);
+      //by default, display all nodes (p-value <= .05)
+      //match position on slider to that of default value
+      document.getElementById('enrichment-selection-slider').value = 0.05;
```
maxkfranz: It would be preferable to set the value as an attribute in render().  React allows a defaultValue, if you only need it set initially.

src/client/features/enrichment/enrichment-menu.js
```diff
@@ -26,7 +26,7 @@ class EnrichmentMenu extends React.Component {
 
     const slider = [
       //set min = 0.0001 to prevent all nodes from being hidden which will occur if min = 0
-      h("input",{type:"range",id:'enrichment-selection-slider',min:0.0001,max:0.05,step:0.0001,defaultValue:0.025,
+      h("input",{type:"range",id:'enrichment-selection-slider',min:0.0001,max:0.05,step:0.0001,defaultValue:0.05,
```
maxkfranz: The defaultValue is set here so the previous DOM manipulation may be redundant

CaitlinOCallaghan:
Setting the slider's defaultValue in the enrichment-menu.js file properly sets the value of the slider; however, the slider's thumb is not updated when re-rendering and its position does not match the value of 0.05.  Both the position and value match when manually manipulating the DOM in base-network-view.js .  I'm not sure why the defaultValue does not set the value and thumb on re-render - it's odd. I can make an issue for it.

CaitlinOCallaghan:
I made updates so that the base-network-view is no longer manually manipulating the DOM.

image


image

jvwong:
When I set the slider to '0' there's still nodes lying around. Is this by design?

src/client/features/enrichment/enrichment-menu.js
```diff
@@ -4,7 +4,37 @@ const { Tab, Tabs, TabList, TabPanel } = require('react-tabs');
...
+        node.addClass('hidden');
+      else
+        node.removeClass('hidden');
+    });
+  }
+
   render(){
+
+    const slider = [
+      //set min = 0.0001 to prevent all nodes from being hidden which will occur if min = 0
```
jvwong: OK I see you've fixed this min to 1E-03. That could get weird if all the nodes are < 1E-08. That kind of adjusted p-value is totally meaningless, but I digress.

Is it weird to fix the range based upon the data? or is that a low priority 'enhancement' we can punt to the next issue/milestone?

CaitlinOCallaghan: I thought about dynamic scaling, but I feel that it's for the next issue/milestone - implementation will get messy because the slider feature uses enrichment.css for the legend background, enrichment-stylesheet.js for the node color, and the .js files in the enrichment feature folder to render the slider

jvwong: OK leave for another day.

PR #918, opened 2018-07-19
Fix gene card link

q4wen:
refs: #856


maxkfranz:
Seems OK when I tried it

jvwong:
ok with me.

Issue #861, opened 2018-06-29
Moving interactions edge weighting & filtering server side


jvwong:
//TODO

maxkfranz:
Reminder from our discussion the other day: Start with doing all the logic on the client to keep things simple to start.  At a later point, it's easy to move the code from client to server, because cytoscape.js code can run in either env.

maxkfranz:
Ref : Client-side network filtering tasks #859

PR #928, opened 2018-07-23
Moved Slider Centrality Calculation server-side

MichaelWrana:
Refs #861 

Also some performance improvements & bug fixes.


MichaelWrana:
400-500ms for just the server-side stuff

src/server/graph-generation/interaction/index.js
```diff
@@ -65,11 +67,50 @@ function parse(data, queryIds){
...
+    const min = Math.min(...centralityVals);
+
+    if(max !== min)
+      centralityMap.forEach( ele => {
+        ele[0].data('bcVal',normalizeValue(ele[1],max,min));
+      });
+    else
+      centralityMap.forEach( ele => {
+        ele[0].data('bcVal',ele[1]);
+      });
```
maxkfranz: Let's make the calculation more abstract.

(1) Do the calculation in a different function, addMetric().
(2) The code that uses parse() can call both parse() and then addMetric().
(3) Don't normalise the values yourself.  The algorithm can do it.  Use the normalised variant.
(4) Store the value under the node in a key called something general like `metric` in case we change the actual value used as the metric

MichaelWrana:
Update: It now takes around 1000ms to filter the network down to 50 nodes & remove all the extra edges

jvwong:
> Update: It now takes around 1000ms to filter the network down to 50 nodes & remove all the extra edges

To be clear, this is on the data from /pc2/graph web service.

maxkfranz:
Step (1) : Switch over to use node.degree() as the metric

Step (2) : Remove cy init.  You don't need a cy instance if you just need to calculate degree.  The parsing already iterates over each interaction, so the degree can just be incremented in-place in the json.

maxkfranz:
Step (3) : Optimise parse and filter etc.

Step (4) : Switch to directly use a metric provided in the extended sif by pc -- maybe centrality or page rank

Step (5) : Make a PC service to directly get cy json, converted from the extended sif.  Then the interaction viewer only needs the fetch() without any post-processing.

maxkfranz:
@MichaelWrana Let's handle (1) and (2) in this PR and we can consider the next steps in other PRs.

IgorRodchenkov:
Regarding Max's very good point 5), and also the centrality metrics, - yes, ideally, PC service (both PC2 and Sifgraph) could return a Cy JSON with app-ui specific metadata/data in there. Let's just discuss and define what we want exactly before rushing into coding it (I wanna see the JSON schema and examples fixed).

Because, It does not sound to me like a good idea to implement and maintain an unofficial and loose output format, e.g., called as _CyJson_ (not a perfect name probably), inside the PC web services in addition to the TXT (aka Extended SIF). Also, Sifgraph service does not have access to all the PC data and metadata that PC2 has (might I could merge cpath2/pc2 and sifgraph-server codebases into one to make this possible...) 

If calculating the centrality or degree metrics alone takes ~60 sec per request, then it won't probably help if we'd move it to the java servers side...

Perhaps, the way to go for the app-ui is still to wisely mix up existing core PC web services, multiple queries, in a desired way, instead of overthinking a new format and how to generate that?..

maxkfranz:
Yes, steps (3), (4), and (5) will need more consideration.  For example, if we decide that degree alone is a good enough metric and the parsing code becomes sufficiently optimised, then we may not do (4) or (5) at all.

We're focussing on implementing just (1) and (2) for now.

Thanks for your feedback, Igor.

MichaelWrana:
I combined `addMetric` and `filterNetwork` into a single function.  Since creating the cytoscape object is needed for `filterNetwork`, it made sense to use that instance to calculate `ele.degree()` rather than calculating degree in the `parse` function.

Note: Even with this method I'm still getting filtering times of 90s for TP53.  Most other nodes I tried are 1-5s.  Maybe we need some special code to handle TP53?

IgorRodchenkov:
TP53 is not the only one that tough. Try NOG.

IR

MichaelWrana:
I got some info on filter time / node number: [RuntimeData.xlsx](https://github.com/PathwayCommons/app-ui/files/2236391/RuntimeData.xlsx)

We probably need something 'quick-and-dirty' to cut it down to ~1000 nodes, then apply the degree filter.

maxkfranz:
I think 4000ms to run the filtering is still too long.  It might be best to just use the degree as the filter for now, Step (1).  See https://github.com/PathwayCommons/app-ui/issues/934#issuecomment-408232911

Step (2) will address most the overhead remaining from (1).

MichaelWrana:
Here is the new timing with step (2).  Seems too quick, like I'm missing something.

NOG:
timings

TP53:
screen shot 2018-07-31 at 12 27 05 pm

If someone else could test, would be super helpful

src/server/graph-generation/interaction/index.js
```diff
@@ -56,20 +64,85 @@ function parse(data, queryIds){
...
+  console.timeEnd('cy');
+
+  console.time('Filter Edges');
+  const filteredEdges = [];
+  edges.forEach( edge => {
+    const source = edge.data.source;
+    const target = edge.data.target;
+    let sourceFound = false;
+    let targetFound = false;
+    filteredNodes.forEach( node => {
```
maxkfranz: You could use a map rather than iterating over all nodes.  That's O(log n) rather than O(n).

src/server/graph-generation/interaction/index.js
```diff
@@ -56,20 +64,85 @@ function parse(data, queryIds){
...
+
+  console.time('Add Edges');
+  filteredEdges.forEach( edge => {
+    try{
+      cy.add({
+        group: 'edges',
+        data: edge.data,
+        classes: edge.classes,
+      });  
+    }catch(err){ return; }
```
maxkfranz: Why try-catch?

src/server/graph-generation/interaction/index.js
```diff
@@ -56,20 +64,85 @@ function parse(data, queryIds){
     edges:[],
     nodes:[],
   };
-  let nodeMap=new Map(); //keeps track of nodes that have already been added
+  let nodeDegreeMap =new Map(); //keeps track of nodes that have already been added, also calculates node degree
```
maxkfranz: What about just having an id => node map?  You can just use node.data.metric directly

maxkfranz:
> Seems too quick, like I'm missing something.

Is it returning the expected results?  Try comparing with the result when filtering is turned off.

MichaelWrana:
TP53 Timings with max's suggestions:

screen shot 2018-08-01 at 12 05 56 pm

Everything looks fine on the network view, I was just surprised it went so fast since we were dealing with 4000ms times before and it's down to 10-15ms now.

maxkfranz:
@MichaelWrana 

Next steps for merge into dev:

(1) Remove console lines

(2) Update your branch with the latest code from the development branch.  There are some conflicts.  VSC should show the conflicts when you try to merge dev into your branch.

PR #931, opened 2018-07-24
Collapse input box #917

CaitlinOCallaghan:
- Click anywhere on screen to close input box
- Synchronize the opening and closing of unrecognized token feedback box
- Open input box (and unrecognized box if invalid tokens exist) when network is loaded

***Remaining issue: clicking/trying to highlight the unrecognized tokens box will cause the input box to lose focus, and, as a result, both boxes close.

### DIFF
--- a/src/client/features/enrichment/index.js
+++ b/src/client/features/enrichment/index.js
@@ -128,7 +128,7 @@ class Enrichment extends React.Component {
   }
 
   render() {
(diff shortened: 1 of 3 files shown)


CaitlinOCallaghan:
As mentioned in the initial comment, there is a bug with this pull request when trying to highlight or copy the list of unrecognized tokens. I didn't want to dwell on it too much, since the feedback mechanism will eventually change and it's not a major functionality issue (its just frustrating for the user).

CaitlinOCallaghan:
I altered it to only open the boxes onLoad if invalid tokens are present. This will act as an indicator to the user to fix their input.

Issue #916, opened 2018-07-19
Using SIFGraph web services 


jvwong:
## Description of new feature

Use the SIFGraph web services for interactions data.
http://www.pathwaycommons.org/sifgraph/swagger-ui.html

q4wen:
New service does not contain node metadata (participant type, uniprot/hgnc id).
Old service:
http://www.pathwaycommons.org/pc2/graph?cmd=graph&source=TP53&pattern=controls-phosphorylation-of&pattern=in-complex-with&pattern=controls-expression-of&pattern=interacts-with&kind=neighborhood&format=txt&user=app-ui
![screen shot 2018-07-24 at 4 17 12 pm](https://user-images.githubusercontent.com/16858225/43163765-220efd82-8f5d-11e8-8c91-b17e65b99555.png)

New service: 
http://www.pathwaycommons.org/sifgraph/v1/neighborhood?direction=BOTHSTREAM&limit=1&pattern=CONTROLS_EXPRESSION_OF&source=TP53

q4wen:
Some data is missing in the new SIF graph service:
1. node metadata (as mentioned in last comment)
2. `pubmedIds` of each interaction
3. Different interaction graph. For example, for interaction graph of TP53 and MDM2,
Old service: 
http://www.pathwaycommons.org/pc2/graph?cmd=graph&source=TP53&source=MDM2&pattern=controls-phosphorylation-of&pattern=in-complex-with&pattern=controls-expression-of&pattern=interacts-with&kind=pathsbetween&format=txt&user=app-ui
![screen shot 2018-07-26 at 12 10 14 pm](https://user-images.githubusercontent.com/16858225/43274326-ee4e0ea4-90cc-11e8-9a65-d04b60b9007d.png)


 New service:  
 http://www.pathwaycommons.org/sifgraph/v1/pathsbetween?directed=false&limit=1&pattern=CONTROLS_PHOSPHORYLATION_OF&pattern=IN_COMPLEX_WITH&pattern=CONTROLS_EXPRESSION_OF&pattern=INTERACTS_WITH&source=TP53&source=MDM2
![screen shot 2018-07-26 at 12 09 01 pm](https://user-images.githubusercontent.com/16858225/43274276-c42493e6-90cc-11e8-9422-2f739dd0fe8a.png)
<br/>

Am I requesting the SIF graph service correctly? I am not sure about the parameters (directed, limit and pattern). @jvwong @IgorRodchenkov

[...]

PR #938, opened 2018-07-26
New SIF Interaction Service

q4wen:
Refs: #916 

Old PC2 graph service is replaced with new SIF graph service. The request returns Json network to the client in 2s (tested with TP53).

### DIFF
--- a/src/server/graph-generation/interaction/index.js
+++ b/src/server/graph-generation/interaction/index.js
@@ -65,13 +65,14 @@ function parse( data, queryIds ){
     const nodeMetadata = null; //TODO: new service does not have node metadata
     data.trim().split('\n').forEach( line => {
       const splitLine=line.split('\t');
-      const nodeA = splitLine[0];
-      const nodeB = splitLine[2];
+      const nodes = [splitLine[0], splitLine[2]];
+      //const nodeA = splitLine[0];
+      //const nodeB = splitLine[2];
       const edge = splitLine[1];
       const mediatorIds = splitLine[4];
 
       const edgeMetadata = interactionMetadata( mediatorIds );
-      addInteraction( [nodeA, nodeB], edge, edgeMetadata, network, nodeMap, nodeMetadata, queryIds );
+      addInteraction( nodes, edge, edgeMetadata, network, nodeMap, nodeMetadata, queryIds );
     });
     return network;
   }
--- a/src/server/graph-generation/interaction/index.js
+++ b/src/server/graph-generation/interaction/index.js
@@ -1,5 +1,4 @@
 const _ = require('lodash');
-const pc = require('../../pathway-commons');
 const logger = require('./../../logger');
 const LRUCache = require('lru-cache');
 const cache = require('../../cache');
@@ -7,7 +6,7 @@ const { PC_CACHE_MAX_SIZE } = require('../../../config');
 const qs = require('query-string');
 const fetch = require('node-fetch');
 
-const neighborhoodUrl = 'http://www.pathwaycommons.org/sifgraph/v1/neighborhood?direction=BOTHSTREAM&limit=1&pattern=CONTROLS_PHOSPHORYLATION_OF&';
+const neighborhoodUrl = 'http://www.pathwaycommons.org/sifgraph/v1/neighborhood?direction=BOTHSTREAM&limit=1&';
 const pathsbetweenUrl = 'http://www.pathwaycommons.org/sifgraph/v1/pathsbetween?directed=false&limit=1&';
 
 function edgeType(type) {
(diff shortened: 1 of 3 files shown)


src/server/graph-generation/interaction/index.js
```diff
@@ -19,23 +23,24 @@ function edgeType(type) {
...
     source : geneIds,
-    pattern : ['controls-phosphorylation-of','in-complex-with','controls-expression-of', 'interacts-with'],
-    kind : geneIds.length > 1 ? 'pathsbetween' : 'neighborhood',
-    format : 'txt'
+    pattern : [ 'CONTROLS_PHOSPHORYLATION_OF','IN_COMPLEX_WITH','CONTROLS_EXPRESSION_OF', 'INTERACTS_WITH' ]
   };
 
-  //Fetch graph from PC
-  return pc.query(params).then(res => {
+  const url = ( geneIds.length > 1 ? pathsbetweenUrl : neighborhoodUrl ) + qs.stringify( params );
```
d2fong: - you should be using pc.query

src/server/graph-generation/interaction/index.js
```diff
@@ -19,23 +23,24 @@ function edgeType(type) {
...
+  return fetch(url)
```
d2fong: All you should be doing is changing the params object to pc.query.

src/server/graph-generation/interaction/index.js
```diff
@@ -51,67 +56,76 @@ const getInteractionGraphFromPC = cache(rawGetInteractionGraphFromPC, pcCache);
...
   };
-  let nodeMap=new Map(); //keeps track of nodes that have already been added
-  if(data){
-    const dataSplit=data.split('\n\n');
-    const nodeMetadata= new Map(dataSplit[1].split('\n').slice(1).map(line =>line.split('\t')).map(line => [line[0], line.slice(1) ]));
-    dataSplit[0].split('\n').slice(1).forEach(line => {
+  let nodeMap = new Map(); //keeps track of nodes that have already been added
+  if( data ){
+
+    const nodeMetadata = null; //TODO: new service does not have node metadata
```
d2fong: @IgorRodchenkov  Can you coordinate with Rex on the missing metadata issue?

src/server/graph-generation/interaction/index.js
IgorRodchenkov: unless you've refactored much there, pc.query uses PC_URL and has nothing to do with the new service (sifgraph); we in fact want to use both PC2 (/get, /traverse, /graph) and Sifgraph endpoints in the app.

IgorRodchenkov:
@d2fong, @q4wen @jvwong 
Wait for me please... I will look into this on Monday.  
If we switch to the sifgraph service, let's have a new method in the pc module rather than messing with pc.query, which is still useful, good to have (for get, traverse queries).

src/server/graph-generation/interaction/index.js
maxkfranz: The pc module can accept any command and query options, right?  They're just parameters.  The concern with using a separate fetch, as this PR does, is that it's duplicated code and it doesn't accept the PC URL env var.

I think we should be using the existing pc module with a command like 'sifgraph/v1/neighborhood'.

IgorRodchenkov:
No, wrong. Look at the code.

IR

IgorRodchenkov:
My point is that PC is not a single server or endpoint (it could be, but not necessary); pc related services can have different api and urls, so why you are trying to hack to squeeze everything into one pc.query method? We don’t do the same for other services, right?

IR

src/server/graph-generation/interaction/index.js
```diff
@@ -51,67 +56,76 @@ const getInteractionGraphFromPC = cache(rawGetInteractionGraphFromPC, pcCache);
...
+    data.trim().split('\n').forEach( line => {
       const splitLine=line.split('\t');
-      const edgeMetadata = interactionMetadata(splitLine[6],splitLine[4]);
-      addInteraction([splitLine[0], splitLine[2]], splitLine[1], edgeMetadata, network, nodeMap, nodeMetadata, queryIds);
+      const nodes = [splitLine[0], splitLine[2]];
+      const edge = splitLine[1];
+      const mediatorIds = splitLine[4];
```
jvwong: I believe this should be `const mediatorIds = splitLine[6];`

src/server/graph-generation/interaction/index.js
```diff
@@ -51,67 +56,76 @@ const getInteractionGraphFromPC = cache(rawGetInteractionGraphFromPC, pcCache);
...
+
+      const edgeMetadata = interactionMetadata( mediatorIds );
```
jvwong: You're missing the second parameter, which should be `splitLine[4]` (see above).

maxkfranz:
> My point is that PC is not a single server or endpoint (it could be, but not necessary); pc related services can have different api and urls, so why you are trying to hack to squeeze everything into one pc.query method?

The point I was trying to get across was that the code for the queries should be re-usable.  If your view of the existing pc module is not compatible with the new sif queries, then perhaps different re-usable code is needed.

> We don’t do the same for other services, right?

If a set of services from a single provider all have similar structure, then I would argue that the commonality would naturally lead to a single module for that provider.  You are more familiar with the pc services than I am, so you would be a better judge.  I see two possibilities:

(1) Add to the existing pc module, perhaps adding a parameter for the "service group", e.g. "sifgraph/v1".

(2) Add a new pc-sif module.

I don't really have a preference either way.  My main concerns are that the code is re-usable and that the base URL is configurable, like `PC_URL` is.

IgorRodchenkov:
Why "a new pc-sif module"? It's fine to add new functions to existing server/pathway-commons module, I think.

src/server/graph-generation/interaction/index.js
IgorRodchenkov: Yeah, I somewhat agree with Dylan and Max; but in order to make it work with like commands cmd:'sifgraph/v1/neighborhood', we have to change the meaning and default value of PC_URL - exclude "pc2/" from it - and do some refactoring (e.g., find/replace 'get' with 'pc2/get', 'search' with 'pc2/search', etc.) through app-ui code.

PR #947, opened 2018-07-31
Expand collapse cue

d2fong:
- adds expand collapse cue for each node on hover.  
- uses tippy tooltips to implement the cue
- 'meta edges' from the expand collapse library are dashed to communicate that they are 'meta edges'
- many subtle complexities and edge cases are handled but there are probably more that I can't think of right now.

screen shot 2018-07-31 at 10 28 40 am
screen shot 2018-07-31 at 10 28 30 am
screen shot 2018-07-31 at 10 28 25 am

### DIFF
--- a/src/client/features/pathways/cy/events.js
+++ b/src/client/features/pathways/cy/events.js
@@ -103,6 +103,7 @@ let bindCyEvents = cy => {
 
   cy.on('mouseover', '.cy-expand-collapse-collapsed-node', nodeHoverExpandCollapse);
 
+  cy.on('mouseout', '$node > node', () => hideCues());
   cy.on('drag', () => hideCues());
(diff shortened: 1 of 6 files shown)


d2fong:
- also it doesn't work on mobile

maxkfranz:
For mobile, one possibility is to show the cue after tapping the node -- since you don't have hover on mobile.

jvwong:
One general observations: The nodes that aren't being expanded/collapsed are being shifted around, sometimes off screen. I'd expect it to be static.

d2fong:
thats a laser cat cose-bilkent thing that makes it look good.

d2fong:
I'm just gonna merge this now, if we find out that we don't want cues down the road we can just remove them.

PR #948, opened 2018-07-31
Changes in pathway-commons module (pc) and where it's used.

IgorRodchenkov:
Refs #916, #938.

PC_URL now must point to the root PC URL (domain or ip:port), such as 'http://www.pathwaycommons.org/' or 'http://beta.pathwaycommons.org/' (tailing slash is required).

This allows app-ui code to query PC using "commands" (request paths) like 'pc2/search' instead 'search', and - now - 'sifgraph/v1/neighborhood', etc. (other/future PC services).


src/server/pathway-commons/index.js
```diff
@@ -44,14 +44,14 @@ const _processQueryString = async (inputString) => {
   ];
 };
 
-//Pathway Commons HTTP GET request; options.cmd = 'get', 'search', 'traverse', 'graph', etc.
+//Pathway Commons HTTP GET request; options.cmd = 'pc2/get', 'pc2/search', 'pc2/traverse', 'pc2/graph', etc.
 const query = async (queryObj) => {
```
jvwong: Is it possible to remove any mention of `'pc2'` especially in the client (e.g. `ServerAPI.pcQuery('pc2/get',`) in favour of appending the 'pc2' here? This would require a change to router for `'/pc2/:path'` to use `query`

src/server/routes/pc.js
```diff
@@ -10,12 +10,17 @@ router.get('/datasources', function (req, res) {
...
 router.get('/querySearch', function (req, res) {
-  pc.querySearch(req.query).then(r => res.json(r));
+  pc.search(req.query).then(r => res.json(r));
 });
 
 //this is mainly to download and save raw data (with correct content-time) from PC ws directly to client
-router.get('/:path', function (req, res) {
-  res.redirect(config.PC_URL + req.params.path + '?' + qs.stringify(req.query));
+router.get('/pc2/:path', function (req, res) {
+  res.redirect(config.PC_URL + 'pc2/' + req.params.path + '?' + qs.stringify(req.query));
```
jvwong: This might be looking for the 'pc2' stripping it, then adding back?. Also could this just point to pc.query method?

IgorRodchenkov:
Try and see it for yourself. It’s not obvious that app-ui calls to download raw data from PC2 work somewhat differently from other pc requests...

Try, try, try... What I’ve done in this PR  doesn’t break the app; further and deeper refactoring and redesign is another matter, which wasn’t my intention this time.

IR

src/server/pathway-commons/index.js
IgorRodchenkov: no

src/server/routes/pc.js
IgorRodchenkov: initially, router.get('/:path', function (req, res) - was to match/route ALL other possible direct requests to PC2 - mainly (or only) for the GSEA/SIF/SBGN/.. download links in the view's right-side panel.
I just had to made this change after changes in pc (server/pathway-commons) module, which redefined what 'cmd' (command) means - now it means full context path, such as pc2/search or sifgraph/v1/pathsbetween

jvwong:
I believe the interaction graph call to pc.query in [/src/server/graph-generation/interaction/index.js](https://github.com/PathwayCommons/app-ui/blob/0dc0e4e6e8c604389cf39f7ede7254f204fc41d3/src/server/graph-generation/interaction/index.js#L26) needs the 'pc2' added .

IgorRodchenkov:
Yes, let me add that (but I think that index.js will be soon refactored via the other pool request - switching to Sifgraph - right?)

jvwong:
Yeah when this goes in then @q4wen can add his from #938

IgorRodchenkov:
Merged. Let me keep igor_pc branch (please don't delete) for future work.

IgorRodchenkov:
Alright, I can live without igor_pc branch :))

Issue #954, opened 2018-08-03
Cache: Case when cache sets value as error


jvwong:
## Environment info

- Browser version (client) : n/a
- Node.js version (server) : n/a

## Current (buggy) behaviour

There is in-memory caching for fetching SIF and [validation](https://github.com/PathwayCommons/app-ui/blob/f6754e6137890a465c007ab1858523cf20b976a5/src/server/enrichment/validation/index.js#L144-L146). In the latter case, I noticed searching for one of the examples 'TP53 MDM2' failed, and each subsequent call thereby returned nothing/Error for that query. 

## Desired behaviour

Don't want failed responses to be cached.

## Minimum steps to reproduce

Difficult to reproduce but can turn off g:Convert call temporarily.

Maybe caching is bad here, remove.?

maxkfranz:
The cache should never store failures.  There should be a `.catch()` that sends an error to the client and leaves the cache as-is.

maxkfranz:
`.catch()` -- on a promise object

PR #957, opened 2018-08-07
reset cache on error

jvwong:
Related to #954: This resets the cache on any error. Otherwise not sure of simple way to get failed key cleanly.


src/server/enrichment/validation/index.js
```diff
@@ -102,6 +103,8 @@ const bodyHandler = body =>  {
  */
 const errorHandler = ( error, query ) => {
   logger.error( error );
+  // something fishy - blow out the cache
+  pcCache.reset();
```
maxkfranz: Pretty sure that would delete everything.  It would be better to delete only the specified key, or even better to leverage the promise chain.

maxkfranz:
Hold off on merging this today, and I'll take a look at this tomorrow.

Issue #958, opened 2018-08-07
Remove logic/operations from service calls to GO/Reactome  


jvwong:
In [/src/client/services/server-api/index.js](https://github.com/PathwayCommons/app-ui/blob/750ba569d987f99bd3445f38485beacb59008042/src/client/services/server-api/index.js#L29-L47), specifically:`getGoInformation` and `getReactomeInformation`:

- log errors 
- input string manipulation tightly coupled to enrichment case ( (i.e. `replace("REAC:...)`)
- response manipulation seems risky (e.g. `res.results[0]...`)

CaitlinOCallaghan:
By 'log errors', do you mean use 'catch' or actually throw an error and log it in the console?

jvwong:
Let's skip the 'logging' - it would be more useful to log IF it was happening on our server.

PR #959, opened 2018-08-07
Remove logic/operations from services #958

CaitlinOCallaghan:
- added error check when determining which service to call -> return promise with value 'null'
- remove logic/operations from service calls
- make service calls generic (manipulate id formatting in enrichment file, rather than in server-api)


src/client/features/enrichment-map/cy/index.js
```diff
@@ -30,11 +30,19 @@ let bindEvents = cy => {
     let node = evt.target;
     let getEnrichmentTooltipData = () => {
       let id = node.data('id');
-      if(id.includes('GO')) return ServerAPI.getGoInformation(id);
-      else if(id.includes('REAC')) return ServerAPI.getReactomeInformation(id);
+      if(id.includes('GO')) return ServerAPI.getGoInformation( id.replace("GO:", "") );
```
jvwong: Safer if these were tests against a regex. I can't find any solid db id information for GO/Reactome, but seem to be similar to `/^GO:\d+$/` etc.

src/client/services/server-api/index.js
```diff
@@ -27,23 +27,15 @@ const ServerAPI = {
   },
 
   getGoInformation(goID) {
-    return fetch(`https://www.ebi.ac.uk/QuickGO/services/ontology/go/search?query=GO%3A${goID.replace("GO:", "")}&limit=1&page=1`, {method: 'GET', timeout: 100})
+    return fetch(`https://www.ebi.ac.uk/QuickGO/services/ontology/go/search?query=GO%3A${goID}&limit=1&page=1`, {method: 'GET', timeout: 100})
```
jvwong: Rethrow any errors for the UI to deal with. See https://github.com/PathwayCommons/app-ui/pull/962#issuecomment-411498990

src/client/features/enrichment-map/cy/index.js
```diff
@@ -30,11 +30,19 @@ let bindEvents = cy => {
...
+      else if(id.includes('REAC')) return ServerAPI.getReactomeInformation( id.replace("REAC:", "R-HSA-") );
+      else return Promise.resolve(null);
     };
 
-    getEnrichmentTooltipData().then( pathwayOverview => {
+    getEnrichmentTooltipData().then( result => {
```
jvwong: Need a `catch()` here and strategy to deal with it.

src/client/services/server-api/index.js
maxkfranz: It's simpler to remove the catch altogether rather than rethrow within the catch.

jvwong: Remove `catch`

PR #962, opened 2018-08-08
Jeff and Max : caching changes and misc. changes

jvwong:
- Update package.json to have `npm run ci` shortcut so it's easy to run the CI tests locally before making a PR.
- Remove cases where promises have their error cases consumed without rethrowing.
- Revise the caching mechanism to selectively remove entries for failed promises.
- Remove enrichment server-side timeout and dummy data returned on failure case.

Supersedes #957

@d2fong Does this make sense with your enrichment-map=>enrichment changes?

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,7 +32,8 @@
     "lint": "run-s lint:*",
     "watch": "run-p watch:*",
     "test": "mocha --recursive",
-    "test:travis": "run-s test lint"
+    "test:travis": "run-s test lint",
+    "ci": "run-s test lint"
   },
   "dependencies": {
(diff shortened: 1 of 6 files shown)


maxkfranz:
@MichaelWrana @q4wen @CaitlinOCallaghan This will affect the enrichment UI and the existing search UI.  Errors are no longer consumed directly in the service functions.  You will see errors in the console when a promise chain is broken by an exception.  The UIs should `.catch()` the promise chains that may have errors, within the UI code itself.  Let's create separate issues for each of these error cases as they come up.

PR #966, opened 2018-08-08
Make the interactions viewer button available when `NODE_ENV=development`

maxkfranz:
- Show the button to open the interactions viewer only when in dev mode (e.g. watch).
- Fix : Clean up the entity list sent through the URL: Just use a simple comma-separated string.

Note: The button would not be visible on appsbeta.pathwaycommons.org.  Another flag would be needed independent of dev/prod for that.  This should suffice for local development of the interactions viewer, and the flag check can be removed completely when it's ready to be released.


maxkfranz:
@jvwong @MichaelWrana @q4wen

Issue #978, opened 2018-08-09
Submitting empty query throws TypeError


jvwong:
## Environment info

- Browser version (client) : n/a
- Node.js version (server) : n/a

## Current (buggy) behaviour

Throws a TypeError.

## Desired behaviour

Deal with this as a valid case.

## Minimum steps to reproduce

Just click submit upon load.

Console:
![image](https://user-images.githubusercontent.com/4706307/43918012-cf150cea-9bdf-11e8-8b95-15b6a9509f14.png)

 Hints:

>src/client/features/enrichment/token-input.js
```
//call validation service API to retrieve validation result in the form of []
  retrieveValidationAPIResult(){
    let { inputBoxContents } = this.state;
    let { controller } = this.props;

    let tokenList = _.pull(inputBoxContents.split(/\s/g), "");
     ServerAPI.enrichmentAPI({
       genes: tokenList,
       targetDb: "HGNCSYMBOL"
      }, "validation")
    .then( result => {
      const aliases = result.geneInfo.map( value => {
        return value.convertedAlias;
      });

      controller.handleGeneQueryResult( {
        genes: aliases,
        unrecognized: result.unrecognized,
      });
    });
  }
```

CaitlinOCallaghan:
Closed

PR #987, opened 2018-08-09
handle empty input #978

CaitlinOCallaghan:
- do not send empty query to service
- if input contains only spaces and line breaks, do not send to service, 
- if input contains only spaces and line breaks, clear input box
- only send inputs with textual content

### DIFF
--- a/src/client/features/enrichment/token-input.js
+++ b/src/client/features/enrichment/token-input.js
@@ -56,7 +56,14 @@ class TokenInput extends React.Component {
           })
         ]),
         h('submit-container', {
-          onClick: () => { this.retrieveValidationAPIResult();} },
+          onClick: () => {
+            if(!inputBoxContents) return;
+            else if(!inputBoxContents.replace(/\s/g, "")) {
+              this.setState({ inputBoxContents: "" });
+              return;
+            }
+            else this.retrieveValidationAPIResult();
+          }},
           [h('button.submit', 'Submit')]
         )
     ]);
--- a/src/client/features/enrichment/token-input.js
+++ b/src/client/features/enrichment/token-input.js
@@ -24,6 +24,11 @@ class TokenInput extends React.Component {
     let { inputBoxContents } = this.state;
     let { controller } = this.props;
 
+    if(!inputBoxContents.replace(/\s/g, "")) {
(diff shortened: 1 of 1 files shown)


src/client/features/enrichment/token-input.js
```diff
@@ -56,7 +56,14 @@ class TokenInput extends React.Component {
           })
         ]),
         h('submit-container', {
-          onClick: () => { this.retrieveValidationAPIResult();} },
+          onClick: () => {
+            if(!inputBoxContents) return;
```
d2fong: I think this logic should go in the retrieveValidationAPIResult code.

d2fong:
Do you think it also makes sense to show a message to the user saying what was wrong with their input.

maxkfranz:
I think the UI message makes sense, and we could add it in some future iteration.  Right now, time may be tight with testing and fixing.

src/client/features/enrichment/token-input.js
```diff
@@ -24,6 +24,12 @@ class TokenInput extends React.Component {
     let { inputBoxContents } = this.state;
     let { controller } = this.props;
 
+    if(!inputBoxContents) return;
```
d2fong: Why does inputBoxContents have to initially come from the parent props?

CaitlinOCallaghan: That's an old part from before the refactoring. I just changed it now since its no longer needed.

d2fong: Ok so you can remove this if statement now

src/client/features/enrichment/token-input.js
```diff
@@ -24,12 +24,6 @@ class TokenInput extends React.Component {
     let { inputBoxContents } = this.state;
     let { controller } = this.props;
 
-    if(!inputBoxContents) return;
-    else if(!inputBoxContents.replace(/\s/g, "")) {
```
d2fong: Sorry I meant that you only had to remove the first part of the else if.

You can return early if the replace check passes.

```js
if(!inputBoxContents.replace(/\s/g, "")) {
   this.setState({ inputBoxContents: "" });	
   return;
}
```

Issue #1000, opened 2018-08-10
Missing nodes/edges


jvwong:
## Current (buggy) behaviour

When viewing an interaction network, moving the slider all the way to 'more' and leaving all the filters off, there are still a number of hidden nodes/edges, the amount of which is different on each page load of the network.

## Desired behaviour

Should see the whole network.

## Minimum steps to reproduce
![image](https://user-images.githubusercontent.com/4706307/43975988-4da5d3a0-9cad-11e8-9120-a1190b630e03.png)

![image](https://user-images.githubusercontent.com/4706307/43976057-6d2f56c4-9cad-11e8-8800-46640154f791.png)

jvwong:
depends on slider rate:
![bug](https://user-images.githubusercontent.com/4706307/43977572-8f9a76b2-9cb2-11e8-81ac-51b4393710d7.gif)

IgorRodchenkov:
I noticed that too, [here](http://appsbeta.pathwaycommons.org/interactions?source=SMAD1).
We should not probably merge PRs before such issues (easy to see) get fixed.

E.g.
![image](https://user-images.githubusercontent.com/991169/43977835-43b02228-9cb3-11e8-8609-e64453618816.png)

d2fong:
Can you see if the branch “slider-click-bug” fixes this issue

jvwong:
Yeah this seems to solve these sets of issues.

Issue #975, opened 2018-08-09
Hiding and showing interactions sidebar resets slider position


MichaelWrana:
## Environment info

- Browser version (client) : Safari latest
- Node.js version (server) :

## Current (buggy) behaviour

Slider position is reset when the interactions sidebar is opened and closed again.

## Desired behaviour

The slider should retain it's previous position after opening and closing the sidebar

## Minimum steps to reproduce
- Open any network in interactions view
- move the slider to a new position
- Close the sidebar and reopen
- Slider will reset position

jvwong:
Also the re-opened slider position is not reflective of the network (vice versa).

MichaelWrana:
Fixed in #1001

Issue #974, opened 2018-08-09
Slider no longer works on 'click'


MichaelWrana:
## Environment info

- Browser version (client) : Safari Latest
- Node.js version (server) :

## Current (buggy) behaviour

In some circumstances, when attempting to click and change the slider rather than slide it, the nodes all disappear, or do not display correctly.

## Desired behaviour
Slider should behave normally when user clicks on a point vs slides it manually.

## Minimum steps to reproduce
- Open any network in interactions
- Click the slider on the right side near "Less"
- Click the slider back to the left side near "More"
- All nodes wil disappear

jvwong:
I see this but it was a little tricky to reproduce:

![bug](https://user-images.githubusercontent.com/4706307/43918761-022e24ca-9be2-11e8-8ad2-99c06f307c25.gif)

jvwong:
In a variation, this query: http://localhost:3000/interactions?source=TP53,ZNF385A results in the network always disappearing and random parts reappearing when the slider is manipulated back and forth.

jvwong:
Possibly related: Setting the slider and the types then returning to original settings results in different network:

![bug](https://user-images.githubusercontent.com/4706307/43967645-dcaf55fa-9c92-11e8-89d8-ce09149dbe4c.gif)

d2fong:
Related #1000

PR #1001, opened 2018-08-10
Slider click bug

d2fong:
This PR should fix all the problems posted in #1000 and #974.  I can't reproduce them in this branch.

## Background info
In some cases the previous iteration of the slider left dangling nodes when you used it. 
My intention was to filter dangling nodes alongside the slider hide operation.

The bug was that it was hiding these dangling nodes in the _same_ operation as the nodes that were being hidden instead of _after_.  I am pretty sure this lead to the behaviour you see in #1000 and #975.  

## Fix
I hide the dangling nodes after adding the class to the nodes to hide.


d2fong:
@maxkfranz @IgorRodchenkov  Thoughts on handling filtering this way compared to your experience implementing this type of functionality in PC viz?

Issue #1010, opened 2018-08-13
genes in tooltip not in alphabetical order


CaitlinOCallaghan:
## Desired behaviour

- alphabetize genes in 'genes shared with entered list' and 'genes in pathway'

![image](https://user-images.githubusercontent.com/38890251/44057526-7a894190-9f19-11e8-8350-ee02518834f9.png)

PR #1011, opened 2018-08-13
Sort gene lists in tooltip #1010

CaitlinOCallaghan:
- alphabetize gene lists in node tooltips to allow user to easily look for a specific gene
image


CaitlinOCallaghan:
I sorted the lists in enrichment-tooltips.js; however, if it would be preferable, I can move the location of .sort() so that the data is stored in the nodes as a sorted array when retrieved from the service calls.

Issue #976, opened 2018-08-09
Some valid genes result in empty networks


CaitlinOCallaghan:
## Environment info

- enrichment app

## Current (buggy) behaviour

some gene queries that are valid produce empty enrichment networks
- ex: "3" (also known as "A2MP1" )

## Desired behaviour

notify user that resulting network is empty

d2fong:
http://localhost:3000/interactions?source=A2MP1

Also doesn't work.

jvwong:
@d2fong I think this is a separate issue?

PR #1013, opened 2018-08-14
empty network view enrichment #976

CaitlinOCallaghan:
- indicate if input resulted in empty network
- allow user to edit input
- adjust placement of message when screen width decreases

image

### DIFF
--- a/src/client/common/components/empty-network.js
+++ b/src/client/common/components/empty-network.js
@@ -9,16 +9,14 @@ class EmptyNetwork extends React.Component {
   render(){
     let { msg, app} = this.props;
 
-    let directions = null;
-    if(app === 'enrichment') directions = h('p','Please edit your gene list');
-    else if(app === 'interactions') directions = ['Return to', h(Link, { className: 'plain-link', to: { pathname: '/' } },  ' Pathway Commons Search')];
+    let linkToPCSearch = (app === 'interactions') ? ['Return to', h(Link, { className: 'plain-link', to: { pathname: '/' } },  ' Pathway Commons Search')] : null;
 
     return h('div.empty-network',[
       h('div.empty-network-header', [
         h('div.pc-logo'),
         h('h1.empty-network-message', msg)
       ]),
-      directions
+      linkToPCSearch
     ]);
   }
 }
--- a/src/client/common/components/empty-network.js
+++ b/src/client/common/components/empty-network.js
@@ -7,16 +7,16 @@ const { Link } = require('react-router-dom');
 
 class EmptyNetwork extends React.Component {
   render(){
-    let { msg, app} = this.props;
+    let { msg, showPcLink } = this.props;
 
-    let linkToPCSearch = (app === 'interactions') ? ['Return to', h(Link, { className: 'plain-link', to: { pathname: '/' } },  ' Pathway Commons Search')] : null;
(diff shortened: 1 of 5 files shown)


src/client/features/enrichment/index.js
```diff
@@ -142,13 +153,21 @@ class Enrichment extends React.Component {
       ])
     ]);
 
+    let emptyNetwork = h('div.empty-enrichment-network', [
```
d2fong: use the empty network component instead 

https://github.com/PathwayCommons/app-ui/blob/development/src/client/common/components/empty-network.js

CaitlinOCallaghan:
Enrichment and interactions empty networks require different user directions (fix gene input VS return to search page). Currently the cases are implemented as if/else - would if be better to use a ternary operator or pass in the directions as a prop instead?

jvwong:
I would think 'No results to display' is sufficiently generic. Also, its probably tough for the user to know how to edit their gene list.

CaitlinOCallaghan:
image

src/client/common/components/empty-network.js
```diff
@@ -7,14 +7,16 @@ const { Link } = require('react-router-dom');
 
 class EmptyNetwork extends React.Component {
   render(){
-    let { msg } = this.props;
+    let { msg, app} = this.props;
+
+    let linkToPCSearch = (app === 'interactions') ? ['Return to', h(Link, { className: 'plain-link', to: { pathname: '/' } },  ' Pathway Commons Search')] : null;
```
d2fong: Instead of passing a string you should pass a boolean prop that determines whether to show the return to pc link or not.

src/styles/common/empty-network.css
```diff
@@ -8,6 +8,10 @@
     width: 5em;
     padding: 20px 0px 0px 0px;
   }
+
+  & .empty-network-message {
```
d2fong: Do you still need this css if you are using the component?

src/styles/features/enrichment.css
```diff
@@ -232,3 +232,26 @@
   box-shadow: 0 1px 1px rgba(0,0,0,0.2);
   border-radius: 5px;
 }
+
+.empty-enrichment-network{
```
d2fong: Do you still need this css if you are using the component?

src/styles/common/empty-network.css
CaitlinOCallaghan: I removed the css from the enrichment.css file since we are using the component.  This part in empty-network.css is to add spacing between the image and the message. I like the view better with the space, but I can remove this portion to. 

image
 VS
image

d2fong: Ok that makes sense

src/client/common/components/empty-network.js
d2fong: i.e. make the prop more general instead of specifically an app string.

In EmptyNetwork:
```js
let { msg, showPcLink } = this.props;

let linkToPCSearch = showPcLink ? ...
```

in enrichment/index.js:
```js

h(EmptyNetwork, { msg: ..., showPcLink: true } )
```

CaitlinOCallaghan: Sorry, I overlooked this comment earlier.

Issue #992, opened 2018-08-09
Enrichment 'search entities' not functional


CaitlinOCallaghan:
## Environment info

- enrichment app

## Current (buggy) behaviour

- 'search entities' box not functional 
- typing in the box results in warnings

![image](https://user-images.githubusercontent.com/38890251/43925131-750d5fee-9bf4-11e8-8e1b-49fa0149484c.png)

## Minimum steps to reproduce

- type a character into 'search entities' box

## Description of new feature

Consider method of searching most valuable to user
- search by node title? 
- search by genes in the node's gene set?

d2fong:
search by genes would be useful but does that mean you would have to send network requests for each node?

PR #1015, opened 2018-08-15
Search entities enrichment functionality #992

CaitlinOCallaghan:
- add functionality to search bar in toolbar 
- allow for searching by node label ('description') and 'geneSet'
image

### DIFF
--- a/src/client/features/enrichment/cy/enrichment-stylesheet.js
+++ b/src/client/features/enrichment/cy/enrichment-stylesheet.js
@@ -88,8 +88,7 @@ const enrichmentStylesheet=cytoscape.stylesheet()
 })
 .selector(`.matched`)
   .css({
-    'border-color': 'yellow',
-    'border-width': 10,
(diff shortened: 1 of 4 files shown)


d2fong:
@maxkfranz @jvwong @q4wen @MichaelWrana thoughts on the highlight colours/functionality?

maxkfranz:
Maybe it would be better to just use the bg colour like the pathway viewer does.  For enrichment and sif cases, you could also make the text outline colour yellow -- that helps with small nodes.

MichaelWrana:
this is a bit much BUT maybe have similar functionality to apple-f on browsers.  Searching a gene would put some arrows below the search bar, then zoom in and centre each node as you click through the list (by clicking the arrows).  Then highlighting could be used to clear up the SBGN view.

jvwong:
Works for me.

CaitlinOCallaghan:
image

Issue #930, opened 2018-07-24
Long text in title extends beyond bounds of app-bar-title container


jvwong:
## Environment info

- Browser version (client) : Chrome 67.0.3396.99 (Official Build) (64-bit)
- Node.js version (server) : n/a

## Current (buggy) behaviour

Long text in titles will extend beyond the app-bar-title container bounds. 

## Desired behaviour

Not extend outside of container.

## Minimum steps to reproduce

See [TP53 regulates transcription of additional cell cycle genes whose exact role in the p53 pathway remain uncertain | Reactome ](http://localhost:3000/pathways?uri=http%3A%2F%2Fidentifiers.org%2Freactome%2FR-HSA-6804115)

![image](https://user-images.githubusercontent.com/4706307/43149869-872c53da-8f36-11e8-97a2-61639b68f81b.png)

d2fong:
Can you still reproduce this?  I can't reproduce it on my computer?

d2fong:
Nevermind, I can reproduce it when I zoom in with Chrome.

jvwong:
I was resizing the browser width (Chrome, Firefox)

PR #1021, opened 2018-08-16
Fit app bar width to contents #930

CaitlinOCallaghan:
- fit width of app-bar to that of its contents

image

### DIFF
--- a/src/styles/features/pathways/app-bar.css
+++ b/src/styles/features/pathways/app-bar.css
@@ -1,5 +1,7 @@
 .app-bar {
-  max-width: 80vw;
+  width: intrinsic;           /* Safari/WebKit uses a non-standard name */
+  width: -moz-max-content;    /* Firefox/Gecko */
+  width: -webkit-max-content; /* Chrome */
   position: absolute;
   top: 0;
   left: 30px;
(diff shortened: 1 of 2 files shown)


src/styles/features/pathways/app-bar.css
```diff
@@ -1,5 +1,7 @@
 .app-bar {
-  max-width: 80vw;
+  width: intrinsic;           /* Safari/WebKit uses a non-standard name */
+  width: -moz-max-content;    /* Firefox/Gecko */
```
d2fong: Generally should not use browser specific css, we pipe all source css to postCSS which should handle this.  Also it seems like intrinsic is experimental; It might not work in older versions of browsers.

MichaelWrana:
Alternate solution: swap the order of elements around  #1023 . Might not be something we want to change, however.

src/styles/features/pathways/app-bar.css
maxkfranz: Shouldn't it suffice to use only `max-width` for specifying width?  I think it's more a matter of cutting off the name if it overflows the width or making the text wrap to the next line.  For wrapping, we would need to consider a max height as well.

PR #1023, opened 2018-08-17
Added links to databases to pathway view title

### LINKED ISSUE #930: Long text in title extends beyond bounds of app-bar-title container

### LINKED ISSUE #1018: Link to pathway and interaction database providers
## Description of new feature

Provide, when available, a link to the pathway or interaction database provider's site.

Specifically, in the pathway title bar. Not sure if there are other instances.

image

## Motivation for new feature

Enhance the navigation to and from data provider sites.

MichaelWrana:
Fixes #1018 
Fixes #930

Reversed the order of `Pathway name | Datasource` => `Datasource | Pathway name`.
Removed "Unknown Title" text if it were to appear.

The map should be correct for all datasources as of right now, but there is also a string similarity calculation included as backup in case the names change in future.

screen shot 2018-08-20 at 12 00 23 pm
screen shot 2018-08-20 at 12 00 38 pm
screen shot 2018-08-20 at 12 00 43 pm

### DIFF
--- a/src/client/features/pathways/index.js
+++ b/src/client/features/pathways/index.js
@@ -4,7 +4,6 @@ const _ = require('lodash');
 const queryString = require('query-string');
 const Loader = require('react-loader');
 const classNames = require('classnames');
-const diceCoefficient = require('dice-coefficient');
 
 const CytoscapeService = require('../../common/cy/');
 const { ServerAPI } = require('../../services/');
@@ -14,7 +13,7 @@ const Sidebar = require('../../common/components/sidebar');
 const EmptyNetwork = require('../../common/components/empty-network');
 const InfoMenu = require('./menus/network-info-menu');
 const FileDownloadMenu = require('./menus/file-download-menu');
-const databaseURLMap = require('../../common/datasource-url-map');
+const datasourceToUrl = require('../../common/datasource-url-map');
 
 const Pathway = require('../../models/pathway/pathway-model');
 
@@ -86,31 +85,10 @@ class Pathways extends React.Component {
     }
   }
 
-  baseNameToUrl(baseName){
-    const lowercaseName = baseName.toLowerCase();
-
-    let databaseURL = databaseURLMap.get(lowercaseName);
-    if(!databaseURL){
-      const datasourceURLScores = [];
-      databaseURLMap.forEach( (datasourceURL, datasource, map) => {
-        datasourceURLScores.push([datasourceURL,diceCoefficient(datasource,lowercaseName)]);
-      });
-      datasourceURLScores.sort( (a,b) => { return b[1] - a[1]; } );
-      databaseURL = datasourceURLScores[0][0];
-    }
-
-    return databaseURL;
-  }
-
   render() {
     let { loading, pathway, cySrv, activeMenu, networkEmpty } = this.state;
 
-    const baseDatasource = pathway.datasource();
-    const databaseURL = this.baseNameToUrl(baseDatasource);
-    let baseName = pathway.name();
-    if(baseName.toLowerCase() === 'unknown title'){
-      baseName = '';
-    }
+    const databaseURL = datasourceToUrl(pathway.datasource());
 
     let network = h('div.network', { className: classNames({
       'network-loading': loading,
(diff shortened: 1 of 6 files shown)


jvwong:
Just curious where did you get the URLs for the databases?

MichaelWrana:
Googled it for most of them.  If I couldn't find a direct link, the link is to the journal article for the database (ex: pid goes to https://www.ncbi.nlm.nih.gov/pmc/articles/PMC2686461/)

jvwong:
OK, I'm fine with the decision for those with no live URL.

MichaelWrana:
Update: Turns out this doesn't actually fix #930

maxkfranz:
How about using `:` instead of `|`?  With the DB name on the left, it looks like the title of the site beside the PC logo.  Maybe a bit more spacing between the DB name and the logo too.

d2fong:
Also, maybe we can make all links consistent and use the plain link style that the tooltips and the protein info box uses

MichaelWrana:
Something like this?

screen shot 2018-08-20 at 10 36 32 am

maxkfranz:
Yes, but I think Dylan meant just the blue underline on links.

d2fong:
There should be a ```plain-link``` class included by the base css file we use.  I think you can just use that.

src/client/features/pathways/index.js
```diff
@@ -84,9 +86,32 @@ class Pathways extends React.Component {
     }
   }
 
+  baseNameToUrl(baseName){
```
d2fong: I think this function belongs in datasource-url-map

src/client/features/pathways/index.js
```diff
@@ -84,9 +86,32 @@ class Pathways extends React.Component {
...
+    return databaseURL;
+  }
+
   render() {
     let { loading, pathway, cySrv, activeMenu, networkEmpty } = this.state;
 
+    const baseDatasource = pathway.datasource();
+    const databaseURL = this.baseNameToUrl(baseDatasource);
+    let baseName = pathway.name();
+    if(baseName.toLowerCase() === 'unknown title'){
```
d2fong: Maybe you can just change the default value returned by pathway.name().  What do you think?

MichaelWrana: I can't figure out where `pathway.name()` is being defined, do you know?

d2fong: https://github.com/PathwayCommons/app-ui/tree/development/src/client/models/pathway

d2fong:
@jvwong thoughts on this?

jvwong:
Works for me!

d2fong:
Thoughts on the way that the name and title are displayed?

jvwong:
Actually, just a sec: Those database names aren't going to work: e.g. pid

jvwong:
'pid' is the only one - probably should be displayed as 'NCI-PID'

MichaelWrana:
I think the names are based on metadata for the network:
In `/src/client/models/pathway`: 
```
  datasource(){
    return _.get(this.raw, 'graph.pathwayMetadata.dataSource.0', 'Unknown datasource');
  }
```

I could add in a quick fix and directly change 'pid' to 'NCI-PID', but that seems like bad practice.  Any ideas?

jvwong:
OK then - I am fine with all of this. 

Notes for future: Datasource derived from [traversal into pathway object](https://github.com/PathwayCommons/app-ui/blob/6e2c136b5f22260335a8b1ac0e57c1c5a224a845/src/server/graph-generation/index.js#L16). Should be improved with [mapping file](https://github.com/PathwayCommons/cpath2/issues/303)

gbader:
Hi - just noticed the move of the database name to be first.  Users don't care about the database name - they want to see the pathway name first.  Database groups, I'm sure, would like to see their name first, but they are secondary to users.  Did you guys discuss that?

d2fong:
I think the initial proposal is in this issue, but I don't think we had any discussion about it.

https://github.com/PathwayCommons/app-ui/issues/1018

jvwong:
That was my suggestion to put the `source: title`- some titles are huge. Anyways,  we can go with `title: datasource`.

Issue #1020, opened 2018-08-16
Example input


jvwong:
## Description of new feature

Provide an example input gene list that will enable to user to 
- understand the type of input accepted
- quickly tour  

## Motivation for new feature

User is told about the app, and doesn't necessarily have data set on hand. Exploratory. Good for quick demos. Determine if the app is appropriate for their needs.

PR #1025, opened 2018-08-17
Enrichment example input #1020

CaitlinOCallaghan:
- example input for enrichment set to: "tyr, oca2, tyrp1, slc45a2"  
(chosen because of quick loading time and variation in node sizes and colours)

- text colour change on hover 'e.g.' 
- 'e.g.' disappears once any text is typed into input box
- input box opens when example input is selected

image
image
Resulting Network
image

### DIFF
--- a/src/client/features/enrichment/token-input.js
+++ b/src/client/features/enrichment/token-input.js
@@ -21,8 +21,8 @@ class TokenInput extends React.Component {
 
   setExample() {
     let example = ['tyr', 'oca2', 'tyrp1', 'slc45a2'];
-    example = example.join(`\n`);
-    this.setState({ inputBoxContents: example });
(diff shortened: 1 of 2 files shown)


jvwong:
Do those genes have any particular meaning?

CaitlinOCallaghan:
The genes are those known to cause albinism. I thought it would be good to use a list where the node names made sense (pigmentation, melanin, vision) with the input. I also considered the network , variation in node sizes and colors, and quick loading speed.

CaitlinOCallaghan:
The input list can be easily changed if we would prefer to use a different set

jvwong:
Oh I just wanted to be prepared for any questions - the results do seem to make sense though!

Issue #1024, opened 2018-08-17
'Close Side Bar' message shows when no side bar


CaitlinOCallaghan:
## Environment info

- all apps  (interactions, enrichment, pathway viewer) - includes live PC pathway viewer

## Current (buggy) behaviour

- 'close side bar' message shows when side bar is closed
![image](https://user-images.githubusercontent.com/38890251/44276588-e97b8780-a215-11e8-93b5-3e61d3ad8afb.png)

## Minimum steps to reproduce
- hover the mouse near the top right corner of any app (interactions, enrichment, pathway viewer)

## Desired behaviour

- do not show 'close side bar' unless side bar is open

PR #1026, opened 2018-08-17
hide 'close sidebar' button #1024

CaitlinOCallaghan:
#1024

- on close, slide the sidebar further to the right to fully hide it
- now if you hover in right corner, nothing shows up

Previously, not all of the sidebar was being hidden, it was just made to be opaque.
Previous view when opacity changed to '1' : 
image


maxkfranz:
When the bar is hidden, could we make it non-interactive altogether as a safeguard, with something like `pointer-events: none`?

Issue #881, opened 2018-07-05
visual consistency in links for the search page


d2fong:
use the plain link visual style for search result titles and all other links.

PR #1029, opened 2018-08-20
change search result titles to use the plain link css class

d2fong:
refs #881


d2fong:
all links on the search page look the same now
screen shot 2018-08-20 at 1 49 22 pm

Issue #990, opened 2018-08-09
Edge tooltips 


jvwong:
## Description of new feature

The interactions and enrichment apps will be able to view metadata related to an edge, including:

- Interactions: navigate and view the SBGN representation (MEDIATOR_IDS) 
- Enrichment: intersecting genes et al   

## Motivation for new feature

Provides more detail regarding on edges 1. from which a 'sif edge' originates (interactions) 2. about the pair of pathways (enrichment).

Issue #967, opened 2018-08-08
Interactions: Consistency in node tooltip 'Links' 


jvwong:
## Description of new feature

In the interactions app, clicking a node reveals a tooltip with relevant metadata, including 'Links' that derive from external database information.

The user should be presented with Links in a more uniform and consistent fashion. For instance, only display links for particular databases, in a location or way that doesn't seem 'random' .

## Motivation for new feature

This will ease navigating the network and tooltip information by having links to external databases in the same location, order.

#### BRCA2
![image](https://user-images.githubusercontent.com/4706307/43860132-bdbdc96a-9b20-11e8-94a6-8710790dd6c4.png)

and the [sif](http://www.pathwaycommons.org/pc2/graph?cmd=graph&source=BRCA2&pattern=controls-phosphorylation-of&pattern=in-complex-with&pattern=controls-expression-of&pattern=interacts-with&kind=neighborhood&format=txt) data:
```txt
BRCA2	DnaReference;ProteinReference;RnaReference	BRCA2;BRCA2 mRNA;BRCA2 promoter;BRCA2_HUMAN;Peptide Linker;brca2_human_gene	ensembl:ENSG00000139618;hprd:02554;ncbi gene:675;refseq:NP_000050;uniprot knowledgebase:P51587	hgnc symbol:BRCA2;intact:EBI-11664513;omim:600185;uniprot knowledgebase:P51587
```

MAP6K
![image](https://user-images.githubusercontent.com/4706307/43860152-c915405e-9b20-11e8-8ca1-91688008c999.png)

and the sif data: 

```txt
MAPK6	ProteinReference	MK06_HUMAN	uniprot knowledgebase:Q16659	hgnc symbol:MAPK6
```

jvwong:
If exists in SIF put it in for now. Optimize later.

q4wen:
![screen shot 2018-08-09 at 11 19 44 am](https://user-images.githubusercontent.com/16858225/43908527-3ebc99d8-9bc6-11e8-9c50-128b05fa6f70.png)

MichaelWrana:
Should probably close this for now considering tooltips are completely disabled on interactions view

jvwong:
I think it's a valid issue - can we reawaken the tooltips?

d2fong:
Should database links be limited to the same ones as the entity info box (uniprot, ncbi gene, gene cards, hgnc)?

jvwong:
[...]

PR #1033, opened 2018-08-21
reawaken Interactions tooltip (node and edge) #967 #990

CaitlinOCallaghan:
- provide functionality for node and edge tooltips
- to close tooltip, click on an element twice, the cy canvas, or a new element (closes first tooltip and opens second)
- all tooltips close when slider is adjusted and when interactions types are filtered
- TODO: fill tooltips with metadata and content

image

image


src/client/features/interactions/interactions-menu.js
```diff
@@ -40,6 +53,8 @@ class InteractionsMenu extends React.Component {
       edges.removeClass('type-hidden');
     }
 
+    this.hideTooltips();
```
d2fong: You should just emit an event here.
i.e.

```js
cy.emit('hide-type');
```
Then, in bindEvents, do something like:
```js
cy.on('hide-type', () => {
  hideTooltips();
});
```

src/client/features/interactions/interactions-menu.js
```diff
@@ -68,6 +70,8 @@ class InteractionsMenu extends React.Component {
       sortedNodes.not(elesToHide).removeClass('metric-hidden');
       elesToHide.addClass('metric-hidden');
     });
+
+    cy.emit('hide-type');
```
d2fong: I think this should be a different event.

Issue #1030, opened 2018-08-21
Handle paths not explicitly declared


jvwong:
## Description of new feature

Provide a sensible redirect/information to a client that navigates to a path that is not explicitly declared.

This will help users get to where they wish to go, or let them know that they navigated to an unknown path/typo. 

## Motivation for new feature

The app should try to inform the client, rather than show a blank page.

PR #1044, opened 2018-08-22
index.html uses root path to get css/js bundles

d2fong:
refs #1043

Currently, all non-api routes render index.html.  The problem when rendering index.html was that the corresponding js and css assets were specified as a relative path.  Leading to the browser trying to get assets like /search/bundle.js when the specified path was /search/.  These assets would not be found leading to the broken behaviour.  

This fix changes the path of the required assets in index.html


d2fong:
Also fixes #1030

jvwong:
Note to self: Probably want to polish this with

- Some navigation item to go somewhere (main search?)
- Use same overall style

PR #1051, opened 2018-08-23
 Enrichment edge tooltips #990

CaitlinOCallaghan:
- edge tooltips for enrichment 
- display intersecting genes in alphabetical order
- added properties `sourceDescription` and `targetDescription` to API to store target and source labels for title - previously, only pathway IDs were being stored
- API documentation updated accordingly for these changes

image
image


jvwong:
These are all useful and good ideas. I'm not convinced that we need to modify the API. Let's keep as a bookmark for later.

Issue #1031, opened 2018-08-21
Show and hiding sidebar cuts off network view port 


jvwong:
## Environment info

- Browser version (client) : Chrome Version 68.0.3440.106 (Official Build) (64-bit); Firefox Quantum 61.0.2 (64-bit)
- Node.js version (server) : n/a

## Current (buggy) behaviour

Hiding the 'info/legend' sidebar cuts off the network where the sidebar once was.

![image](https://user-images.githubusercontent.com/4706307/44407951-5734f580-a52d-11e8-8a9b-5107ae0a54b2.png)

## Desired behaviour

Should be able to view network within the area where sidebar once was.

## Minimum steps to reproduce

Load example network. Close sidebar. Drag network towards right.

jvwong:
Maybe this will be moot if #1012 is implemented.

maxkfranz:
> Maybe this will be moot if #1012 is implemented.

Yes, but we should have a shortterm fix for the upcoming release.

CaitlinOCallaghan:
Should we have the side bar open when the page is loaded? 

This will fix the problem, and also allow users to access the FAQ before trying to input a gene list

![image](https://user-images.githubusercontent.com/38890251/44413817-4ccd2880-a53a-11e8-893b-a78e544a7ebe.png)

CaitlinOCallaghan:
If we do leave the side bar open on initial loading, then I think we should also render the toolbar in case the user closes the side menu. The only thing is that it gets confusing with 2 input/search/typable bars in such close proximity. 

![image](https://user-images.githubusercontent.com/38890251/44413976-a897b180-a53a-11e8-9b5f-5b3710b3762a.png)

maxkfranz:
I wouldn't worry about the confusion part re. the text boxes at this point in time.  As long as the graph spacing issue is addressed, I think it should be OK.

jvwong:
I don’t understand how having the sidebar open or closed affects this bug

CaitlinOCallaghan:
For the network there is a class `'network-sidebar-open'` that controls the width of the view port. This is currently hard coded to true so that the class is always applied even if the sidebar is closed, which is the bug. 

[...]

PR #1056, opened 2018-08-27
Fix changemenu

d2fong:
refs #1031


maxkfranz:
With future refactoring, we may just want to have a generic `Sidebar` component that each view could re-use directly.  But that's probably not necessary if we get rid of the sidebar anyway.

For now, this PR's approach is fine as long as it works.

PR #1059, opened 2018-08-27
create cytoscape network component

d2fong:
- create a cytoscape network component that handles componentWillMount/unMount logic.  
- parent components pass a 'onMount' prop that is called after the network component mounts
- the onMount prop is usually a server call to get some cytoscape json which is then loaded into cytoscape

@maxkfranz thoughts on this?


maxkfranz:
Makes sense

PR #1063, opened 2018-09-10
change phosphorlation type to modification

d2fong:


jvwong:
Are you aware of a query where we can see either 'other' interactions?

d2fong:
I haven't found one yet.

PR #1072, opened 2018-09-14
Tooltips & misc. fixes

maxkfranz:
This PR amends the tooltip functionality.  It also includes some cleanup in code that was touched by the tooltip changes.  Changes are applies to both the SBGN view and the SIF view.

Change list:

- Tooltips do not conflict with popovers anymore.
- Popovers are used for download and info buttons.
- The results page toolbar is cleaned up re. CSS.  It just uses the included styles in base.css for built-in components.
- The search bar takes up the remainder of space to accommodate the floating toolbar.
- Updates Tippy and React dependencies.
- Miscellaneous cleanup.

Preview:

screen shot 2018-09-14 at 4 47 52 pm


maxkfranz:
I don't have Chrome 68, and it seems there's no way to downgrade.

jvwong:
This is also happening on Version 69.0.3497.92 (Official Build) (64-bit)

maxkfranz:
Not on 69.0.3497.81

maxkfranz:
I think bdb4e87 should resolve it for Chrome

jvwong:
Yes that did it.

maxkfranz:
I think they reversed parts of the v69 release in a recent patch (because 69 has many obvious regressions), which is probably why Chrome v69 may behave like v68 for this issue.

maxkfranz:
I'm not getting the 69.x.y.92 patch on my machine for some reason

PR #1073, opened 2018-09-17
Use a MAX_SIF_NODES env var to configure the maximum number of nodes …

maxkfranz:
…returned by the SIF view.  (Also corrects the query node metric to use max int instead of Infinity.

- Sets num to 25.
- Less hairballs.
- Allows configuration by env var for easy testing of different values.
- Improves layout quality.
- Improves layout performance.


maxkfranz:
> Sure sounds good. Future 'service' - would be nice to let client override

Perhaps, but we'd have to plan out how this would interact with the caching.

PR #1076, opened 2018-09-17
Make the SIF viewer link similar to all the other viewer links

maxkfranz:
Since it looks like a normal search result, the user may be more likely to click it.

Preview:

screen shot 2018-09-17 at 12 40 56 pm

screen shot 2018-09-17 at 12 47 30 pm

screen shot 2018-09-17 at 12 47 07 pm

### DIFF
--- a/src/client/features/search/entity-info-box.js
+++ b/src/client/features/search/entity-info-box.js
@@ -88,7 +88,7 @@ class EntityInfoBoxList extends React.Component {
         return `Interactions with ${sources[0]}`;
       }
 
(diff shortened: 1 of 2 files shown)


maxkfranz:
Note: The participant number is excluded for the SIF entry, as that's probably not easily accessible information.

jvwong:
This works for me - we could use outside input on this.

Given how fast that SifGraph service comes back (~300ms for big query) would be fun to  entertain the idea of embedding the result in the future?:

image

maxkfranz:
Eventually, it would be nice to have preview images used for other results as well using the cytosnap lib.

Issue #1070, opened 2018-09-14
interactions loading function fails for some queries


d2fong:
## Desc.
Some interactions queries are really slow:

This one has been spinning for > 2 minutes and it is still going:
http://appsbeta.pathwaycommons.org/interactions?source=ROCK1

## Context
This particular query (ROCK1) was pretty fast before we fetched the other interaction types.
I believe that it is partially due to the layout because TP53 ( a network 3x the size of ROCK1 ) loads quickly.

1.  It loads fast if the layout is not run
2.  It seemingly spins forever when the layout is run inside the loading function
3.  Running the layout outside of this loading function seems to work.

jvwong:
i noticed that the 'controls-state-change-of' bogs things down.

d2fong:
I think it may have to do with the onStop function for cy layouts.  

This doesn't work for ROCK1, but it works for every other example I have tried so far.
```js
      cy.layout(_.assign({}, INTERACTIONS_LAYOUT_OPTS, {
        stop: () => {
          this.setState({
            loading: false,
          });
        }
      })).run();
```

This works for everything
```js
      this.setState({loading: false});
      cy.layout(_.assign({}, INTERACTIONS_LAYOUT_OPTS, {
        // name: 'random',
        stop: () => {
          this.setState({
            loading: false,
          });
        }
      })).run();
    };
```

It makes me believe there is something wrong with the stop() function but that doesn't explain why it only happens for ROCK1

d2fong:
http://appsbeta.pathwaycommons.org/interactions?source=FGF9 Also doesn't load but I am not sure if it is for the same reasons.

jvwong:
Not sure about the layout stuff but I can get FGF9 to load when I drop all but one ('phosphorylation-of') of the patterns on the PC/graph call. 

I think we're hitting the ceiling for what we can do on the app-ui side (caching, reduced pattern types). We all know the graph calls are pretty expensive and likely never designed for this purpose.

maxkfranz:
If the network size is an issue, let's just cap what the server sends back to 25 nodes.  You'll get better layout quality as well, since COSE-Bilkent can't handle the nodes hidden by the slider.  If the server only sends 25, you can just disable the slider entirely and use all the nodes that come back.

PR #1077, opened 2018-09-17
Iss916 - Use SIFGraph service

jvwong:
re #916 and also  #1070

I just assume the nodes are HGNC symbols - this won't fly with CHEBI but I'm restricting patterns right now


maxkfranz:
For the upcoming release, I suppose #1070 is the more pressing issue?  How does the server perform with the new N=25 restriction?  Still too slow?

d2fong:
http://appsbeta.pathwaycommons.org/interactions?source=FGF9 seems ok with 25 nodes
http://appsbeta.pathwaycommons.org/interactions?source=ROCK1 doesn't seem to be affected by network size.

d2fong:
You are going to have to update the tests:
https://travis-ci.org/PathwayCommons/app-ui/jobs/429666928

/test/server/interactions/interactions-network-generation-test.js

test/server/interactions/interactions-network-generation-test.js
```diff
@@ -150,32 +95,4 @@ describe('Interactions network to cy.js conversion', function(){
 
   });
 
-  it('filters nodes with 0 degree', function(){
```
d2fong: I still think it is useful to have a test showing that we filter degree 0 nodes.

maxkfranz: I agree it's better to catch the edgecases as best as possible with the tests

Issue #1082, opened 2018-09-18
Interaction datasource


jvwong:
## Description of new feature

Capture and possibly provide the datasource(s) for the interaction (like PCViz does). Will inform one of data provider(s) and origin. 

## Motivation for new feature

Data providers could appreciate having this type of data available.  Could link to provider if available. Sometimes people are interested in only certain data providers.

PR #1083, opened 2018-09-19
Added datasources but no link

jvwong:
re #1082 adding datasources for interactions but no link.

### DIFF
--- a/src/client/features/interactions/interactions-edge-tooltip.js
+++ b/src/client/features/interactions/interactions-edge-tooltip.js
@@ -1,7 +1,7 @@
 const React = require('react');
 const h = require('react-hyperscript');
 const queryString = require('query-string');
-
+const _ = require('lodash');
 
 const { ServerAPI } = require('../../services');
 
@@ -10,8 +10,17 @@ class InteractionsEdgeTooltip extends React.Component {
     super(props);
 
     this.state = {
+      datasources: [],
       publications: []
     };
+
+    // Update datasources content to get homepage?
+    ServerAPI.datasources()
+      .then(result => {
+        this.setState({
+          dataSources: result
+        });
+      });
   }
 
   componentDidMount(){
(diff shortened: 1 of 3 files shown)


jvwong:
example for tp53 mdm2

image

src/client/features/interactions/interactions-edge-tooltip.js
```diff
@@ -10,8 +10,17 @@ class InteractionsEdgeTooltip extends React.Component {
     super(props);
 
     this.state = {
+      datasources: [],
       publications: []
     };
+
+    // Update datasources content to get homepage?
+    ServerAPI.datasources()
```
d2fong: Why can't you send this request once at the time the network is generated and compute the datasources for all elements there?

src/client/features/interactions/interactions-edge-tooltip.js
```diff
@@ -24,11 +33,22 @@ class InteractionsEdgeTooltip extends React.Component {
...
   render(){
+    const state = this.state;
     let { edge } = this.props;
     let { publications } = this.state;
     let title = edge.data('id');
+    let providers = edge.data('providers');
     let pcIds = edge.data('pcIds');
 
+    let providersList = providers.map( provider => {
+      const dsInfo = _.isEmpty(state.dataSources) ? {}: _.find(state.dataSources, ds => {
```
d2fong: I don't like introducing another instance of the 'find datasource by id' problem into another area of the codebase.  We need to come up with a solution to keep this logic in one place and make it reusable.

src/server/graph-generation/interaction/index.js
```diff
@@ -48,7 +50,9 @@ let interactionTxt2CyJson = (srcId, tgtId, type, pubmedIdsString, mediatorIdsStr
       type,
       source: srcId,
       target: tgtId,
+      providers,
```
d2fong: You need to update the tests.

jvwong:
ok yeah better you than me. I was having a hard time figuring out how to do all that. THanks!

PR #1100, opened 2018-09-27
show sbgn class as a fallback when the node has no metadata

d2fong:
screen shot 2018-09-27 at 4 20 25 pm
screen shot 2018-09-27 at 4 22 20 pm


jvwong:
I'm fine with this. 

Just reiterate the trade-off: Disabling tooltips may confuse user, might think it is 'broken'.

d2fong:
ok

Issue #1105, opened 2018-10-03
Misc User feedback - Pt1


jvwong:
- Indicate interactions displayed
  - Interactions title: 'Top 25 TP53 interactions'
  - Search listing: 'Top 25 TP53 Interactions'

- Provide more informative name for 'Pathway Links'
  - e.g. 'View interactions'
  - Remove square brackets (look like references)  

For later:
- Directly show interactions for peripheral/non-source node
  - Launch new tab vs reload?

PR #1106, opened 2018-10-03
Iss1105 - Misc user feedback

jvwong:
refs #1105 

- interaction listing and title 
- interaction tooltip link outs

screenshot at oct 03 10-30-52
screenshot at oct 03 10-31-20
screenshot at oct 03 10-31-35

### DIFF
--- a/src/client/features/interactions/index.js
+++ b/src/client/features/interactions/index.js
@@ -5,6 +5,7 @@ const queryString = require('query-string');
 const Loader = require('react-loader');
 const classNames = require('classnames');
 
+const config = require('../../../config');
 const CytoscapeService = require('../../common/cy/');
 const { ServerAPI } = require('../../services/');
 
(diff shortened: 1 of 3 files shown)


src/client/features/interactions/index.js
```diff
@@ -72,7 +73,8 @@ class Interactions extends React.Component {
     let appBar = h('div.app-bar.interactions-bar', [
       h('div.app-bar-branding', [
         h(PcLogoLink),
-        h('div.app-bar-title', sources.join(', ') + ' Interactions')
+        sources.length === 1 ?  h('div.app-bar-title', `Top ${config.MAX_SIF_NODES} interactions involving ${sources[0]}`):
```
d2fong: should we always display max sif nodes config value if there are less nodes than that?  e.g. top 25 interactions between mdm2 and tp53 when there are only two actual nodes?

maxkfranz:
It's top 25 genes rather than interactions, right?

How about something like this?

n = 1 : Interactions with TP53 and the top 25 genes

For n = 1, we can have up to 25 resultant genes.

n >= 2 : Interactions between TP53, MDM2, and PCNA

For n >= 2, we only ever return the n specified with interactions between them.

maxkfranz:
An alternative for n = 1 : TP53 and its interactions with the top 25 genes

jvwong:
This one has:

n = 1: 'Interactions between TP53 and top 25 genes'
n >= 2: 'Interactions between TP53, MDM2 and CDKN1A'

maxkfranz:
@jvwong @d2fong I've added some padding to make the link sizes a bit bigger, like they were when they had parentheses.  This makes them easier to click.

d2fong:
You are going to have to cherry pick all of these commits to v2.0.0.  Thats the release branch.

d2fong:
@maxkfranz  @jvwong 

Also don't merge development into v2.0.0.  There is stuff in development that shouldn't go into v2.0.0.

maxkfranz:
Cherry-picked this PR into v2.0.0

Issue #1104, opened 2018-10-02
Refactoring 'info box' data querying


jvwong:
Background: The search 'info box' displays entity data for biological IDs entered. The data fetch process involves tokenization, query prefix tests, ID-mapping and fetching records from external databases (e.g. UniProt, NCBI, HGNC). 

![search_ 2 0](https://user-images.githubusercontent.com/4706307/46365104-b1a19580-c645-11e8-9daf-d85d232afc2c.png)

Issues: The pathway search code perform similar operations, therefore there is much code duplication and redundant calls to external services. Also, the info box query code exists on the client whereas pathway search resides server-side. Finally, the info box query logic could be greatly simplified and modularized.

- Proposed tasks/issues
  - model an 'entity summary', used to populate info box
  - create external service modules for HGNC, NCBI Gene and UniProt to get summary data
  - service for fetching entity summaries from IDs
  - swap out new service to populate UI

Later, unify redundant code for pathway search and fetching entity summary.

jvwong:
Part I:
  - model an 'entity summary', used to populate info box
  - create external service modules for HGNC, NCBI Gene and UniProt to get summary data
  - service for fetching entity summaries from IDs

jvwong:
Part II:
  - ModifyUI to utilize new service

Issue #1108, opened 2018-10-03
evaluate whether we need external-service-data


d2fong:
In /src/client/common/external-service-data.js is a list of databases like CHEBI, KEGG, Gene Cards, etc.

It is currently only used in one place in /src/models/pathway-node-metadata.js to generate urls linkouts to databases that appear in the pathway node metadata.

Requirements:
- find out if we even need all of these databases in this file or if we only need a select few
- move this file to the models folder if we still need it

jvwong:
Let's support link out to the following xref db:
  - 'chebi'
  - 'hgnc symbol'
  - 'uniprot knowledgebase'
  - 'reactome'
  - 'ensembl'
  - 'ncbi gene'

Merge the relevant entries in `common/external-service-data.js` into `models/datasources.js` by adding a 'meta resolver' value, upon which we can append the local identifier. 

Everything will be hardcoded in a file but at least its all in once place.

jvwong:
Also merge `src/models/entity/summary.js` `DATASOURCES`.

d2fong:
To be replaced by the identifers url service

PR #1113, opened 2018-10-15
Moving 'info' data fetch to server

jvwong:
Refs #1104, part one.

- simple model/definition for 'entity summary' 
- external services for fetch from HGNC, NCBI Gene and UniProt 
- service to create summaries from query string (q):
  - example for `q=tp53 is an important gene`

```json
{
	"tp53": {
		"dataSource": "http://identifiers.org/ncbigene/",
		"displayName": "tumor protein p53",
		"localID": "7157",
		"description": "This gene encodes a tumor suppressor protein containing transcriptional activation, DNA binding, and oligomerization domains. The encoded protein responds to diverse cellular stresses to regulate expression of target genes, thereby inducing cell cycle arrest, apoptosis, senescence, DNA repair, or changes in metabolism. Mutations in this gene are associated with a variety of human cancers, including hereditary cancers such as Li-Fraumeni syndrome. Alternative splicing of this gene and the use of alternate promoters result in multiple transcript variants and isoforms. Additional isoforms have also been shown to result from the use of alternate translation initiation codons from identical transcript variants (PMIDs: 12032546, 20937277). [provided by RefSeq, Dec 2016]",
		"aliasName": [
			"cellular tumor antigen p53",
			"antigen NY-CO-13",
			"mutant tumor protein 53",
			"p53 tumor suppressor",
			"phosphoprotein p53",
			"transformation-related protein 53",
			"tumor protein 53",
			"tumor supressor p53"
		],
		"aliasId": [
			"BCC7",
			"LFS1",
			"P53",
			"TRP53"
		],
		"xref": {
			"http://identifiers.org/hgnc/": "TP53",
			"http://identifiers.org/genecards/": "TP53",
			"http://identifiers.org/uniprot/": "P04637"
		}
	}
}
```

### DIFF
--- a/src/models/entity/summary.js
+++ b/src/models/entity/summary.js
@@ -1,0 +1,14 @@
+// Entity summary
+class EntitySummary {
+  constructor( dataSource, displayName, localID, description, aliasName, aliasId, xref ){
+    this.dataSource = dataSource || '';
+    this.displayName = displayName || '';
+    this.localID = localID || '';
+    this.description = description || '';
+    this.aliasName = aliasName || [];
+    this.aliasId = aliasId || [];
+    this.xref = xref || {};
+  }
+}
+
+module.exports = { EntitySummary };
--- a/src/models/entity/summary.js
+++ b/src/models/entity/summary.js
@@ -1,4 +1,7 @@
 // Entity summary
+const _ = require('lodash');
+
+//To be moved and integrated with all the other 'datasource'
 const DATASOURCES = {
   NCBIGENE: 'http://identifiers.org/ncbigene/',
   HGNC: 'http://identifiers.org/hgnc/',
@@ -6,15 +9,19 @@ const DATASOURCES = {
   GENECARDS: 'http://identifiers.org/genecards/'
 };
 
+const DEFAULTS = {
+  dataSource: '',
+  displayName: '',
+  localID: '',
+  description: '',
+  aliases: [],
+  aliasIds: [],
+  xref: {}
+};
+
 class EntitySummary {
-  constructor( dataSource, displayName, localID, description, aliases, aliasIds, xref ){
-    this.dataSource = dataSource || '';
-    this.displayName = displayName || '';
-    this.localID = localID || '';
-    this.description = description || '';
-    this.aliases = aliases || [];
-    this.aliasIds = aliasIds || [];
-    this.xref = xref || {};
+  constructor( options ){
+    _.assign(this, DEFAULTS, options );
   }
 }
 
--- a/src/models/entity/summary.js
+++ b/src/models/entity/summary.js
@@ -1,4 +1,11 @@
 // Entity summary
+const DATASOURCES = {
+  NCBIGENE: 'http://identifiers.org/ncbigene/',
+  HGNC: 'http://identifiers.org/hgnc/',
+  UNIPROT: 'http://identifiers.org/uniprot/',
+  GENECARDS: 'http://identifiers.org/genecards/'
+};
+
 class EntitySummary {
   constructor( dataSource, displayName, localID, description, aliasName, aliasId, xref ){
     this.dataSource = dataSource || '';
@@ -11,4 +18,4 @@ class EntitySummary {
   }
 }
 
-module.exports = { EntitySummary };
+module.exports = { EntitySummary, DATASOURCES };
--- a/src/models/entity/summary.js
+++ b/src/models/entity/summary.js
@@ -7,13 +7,13 @@ const DATASOURCES = {
 };
 
 class EntitySummary {
-  constructor( dataSource, displayName, localID, description, aliasName, aliasId, xref ){
+  constructor( dataSource, displayName, localID, description, aliases, aliasIds, xref ){
     this.dataSource = dataSource || '';
     this.displayName = displayName || '';
     this.localID = localID || '';
     this.description = description || '';
-    this.aliasName = aliasName || [];
-    this.aliasId = aliasId || [];
+    this.aliases = aliases || [];
+    this.aliasIds = aliasIds || [];
     this.xref = xref || {};
   }
 }
--- a/src/server/external-services/hgnc.js
+++ b/src/server/external-services/hgnc.js
@@ -17,8 +17,7 @@ const fetchBySymbols = async symbols => {
   const docs = symbols.map( async symbol => {
     // Process each symbol individually
     const result = await fetchBySymbol( symbol );
-    if( result.response.numFound !== 1 ) return;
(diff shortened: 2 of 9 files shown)


src/server/external-services/hgnc.js
```diff
@@ -0,0 +1,61 @@
...
+  return await Promise.all( docs );
+};
+
+const getEntitySummary = async symbols => {
+
+  const summary = {};
+  if ( _.isEmpty( symbols ) ) return summary;
+
+  const results = await fetchBySymbols( symbols );
+  const nonEmptyResults = _.filter( results, o => _.get( o, 'response.numFound') );
```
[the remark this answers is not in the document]
maxkfranz: It's fine to use async/await on the server like this example, but do not do that on the client.  The polyfill for that is large, and [it's still a relatively new feature](https://caniuse.com/#search=await).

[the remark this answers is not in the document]
jvwong: > You should use array map and array filter instead of lodash map/filter.

What's the difference between the two?

src/server/routes/summary/entity/index.js
```diff
@@ -0,0 +1,79 @@
...
+ * entitySearch: Search a list of strings for recognized gene/protein identifiers
+ * and return an EntitySummary for each
+ *
+ * @param { array } tokens string(s) that should be queried
+ * @return { object } value is EntitySummary keyed by the input token
+ */
+const entitySearch = async tokens => {
+
+  // look for special tokens
+  if ( tokens.length === 1 ) {
```
d2fong: If tokens is not length 1, then this returns undefined.  Maybe you should throw an error or return {}?

src/server/routes/summary/index.js
```diff
@@ -0,0 +1,10 @@
+const express = require('express');
+const router = express.Router();
+const { entitySearch } = require('./entity');
+
+router.get('/entity/search', function (req, res) {
+  let tokens = req.query.q.trim().split(' ');
+  entitySearch( tokens ).then( r => res.json( r ) );
```
d2fong: Not sure what the behaviour is when you res.json( undefined )

src/server/external-services/hgnc.js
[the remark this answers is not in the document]
maxkfranz: `_.filter()` can be more useful than `array.filter()` if you're doing functional programming.  I don't really have a preference anymore, since we're using lodash so extensively overall in this project.

src/server/routes/summary/entity/index.js
jvwong: In the case of token length >1 it just passes right through, doesn't it?

d2fong: Yeah you are right I didn't notice that there is more to this function.

src/server/routes/summary/index.js
maxkfranz: There should be a `.catch()` that then sends an http error code via `res`.

d2fong: @jvwong add error handling and then we can just merge this.

d2fong: If I was the client, I think I would just care about whether the request succeeded or not.  On the server, it might be useful to log the error so that we can know the actual cause of failure.

maxkfranz: It's usually best to not just send the same error object to the client.  In some cases, it could contain private information.  For this case, privacy probably doesn't matter.

The other general consideration is context, like Dylan mentioned.  It's good to summarise the error in plain readable language that accommodates the client's usecase.  

E.g. `res.status(500).send('Failed to access EBI')`

src/server/routes/summary/index.js
```diff
@@ -0,0 +1,12 @@
+const express = require('express');
+const router = express.Router();
+const { entitySearch } = require('./entity');
+
+router.get('/entity/search', function (req, res) {
+  let tokens = req.query.q.trim().split(' ');
+  entitySearch( tokens )
+    .then( r => res.json( r ) )
+    .catch( e => res.status( 500 ).send( { "error": "Server error" } ));  // eslint-disable-line no-unused-vars
```
maxkfranz: I don't think JSON makes sense in the case of an error.  It's going to be plain text.

src/models/entity/summary.js
```diff
@@ -0,0 +1,21 @@
+// Entity summary
+const DATASOURCES = {
+  NCBIGENE: 'http://identifiers.org/ncbigene/',
+  HGNC: 'http://identifiers.org/hgnc/',
+  UNIPROT: 'http://identifiers.org/uniprot/',
+  GENECARDS: 'http://identifiers.org/genecards/'
```
maxkfranz: Should these be in conf?

src/models/entity/summary.js
```diff
@@ -0,0 +1,21 @@
...
+};
+
+class EntitySummary {
+  constructor( dataSource, displayName, localID, description, aliases, aliasIds, xref ){
```
maxkfranz: There are so many args, it's hard to know what is what.

It's better to do this:

```js
constructor({ dataSource, displayName, localID, description, aliases, aliasIds, xref }){
  // ...
}
```

That is, you pass an object whose fields are the args.  That way you can see what arg is what based on the object keys.

maxkfranz: Or just do...

```js
constructor(options){
  _.assign(this, DEFAULTS, options);
}
```

... where you define `DEFAULTS` in the top scope.

src/server/external-services/hgnc.js
```diff
@@ -0,0 +1,66 @@
...
+};
+
+const fetchBySymbols = async symbols => {
+  const docs = symbols.map( async symbol => {
+    // Process each symbol individually
+    const result = await fetchBySymbol( symbol );
+    return result;
+  });
+  return await Promise.all( docs );
+};
```
maxkfranz: This is just one line.  It would be more readable:

```js
const fetchBySymbols = symbols => Promise.all( symbols.map(fetchBySymbol) );
```

src/models/entity/summary.js
jvwong: Yeah we need to put these somewhere global and integrate similar stuff including
  - `src/client/common/external-service-data.js` #1108  
  - `src/models/datasources.js` 

Will deal with as part of #1108.

src/server/routes/summary/index.js
jvwong: Returning text.

src/models/entity/summary.js
jvwong: ok good call. much better...

PR #1115, opened 2018-10-19
rework pathway metadata generation

d2fong:
## Summary
- re implemented pathway metadata generation

## Major changes
- delete pathway metadata model since it was just a convenience wrapper around the original metadata
- slightly simpler implementation  (~150lines (new) vs ~700 lines (old) ) 
- makes it easier to work with metadata ( plain object instead of nested arrays inside nested arrays )
- adds clear input / output test case

## Minor changes
- tries to explicitly declare the hacks used to append biopax metadata into the sbgn json model
- adds a debug command to the package.json to debug server side code from the chrome debugger
- adds an identifiers.org config URL

## TODO
- [ ] review current comments to see if they are descriptive enough to communicate the hacks used
- [x] hgnc linkouts from identifiers.org is currently wrong (```/hgnc symbol``` instead of ```hgnc.symbol```.  
- [ ] there are many instances of hardcoded 'indentifiers.org' values floating around the codebase, all the logic related to generating identifers urls should be put in a module


src/client/common/external-service-data.js
```diff
@@ -1,28 +1,20 @@
-module.exports = [
-  {database:'BioGrid', url:'http://identifiers.org/biogrid/', search:''},
```
d2fong: We should add GO here because I see references to identifiers/go in the enrichment app

src/client/features/paint/pathway-node-metadata.js
```diff
@@ -1,81 +1,120 @@
...
+
+const { generateIdentifiersUrl } = require('../../common/external-service-data');
+const { ServerAPI } = require('../../services');
+
+const DEFAULT_NUM_NAMES = 3;
+const SUPPORTED_DB_LINKS = {
+  'Reactome': 'reactome',
+  'UniProt': 'uniprot knowledgebase',
+  'NCBI Gene': 'ncbi gene',
+  'HGNC Symbol': 'hgnc symbol'
```
d2fong: 'hgnc symbol' is the id currently used to generate identifiers urls but it should actually be 'hgnc.symbol'

src/client/features/paint/pathway-node-metadata.js
```diff
@@ -1,81 +1,120 @@
...
+const _ = require('lodash');
+
+const { generateIdentifiersUrl } = require('../../common/external-service-data');
+const { ServerAPI } = require('../../services');
+
+const DEFAULT_NUM_NAMES = 3;
+const SUPPORTED_DB_LINKS = {
+  'Reactome': 'reactome',
+  'UniProt': 'uniprot knowledgebase',
+  'NCBI Gene': 'ncbi gene',
```
d2fong: 'ncbi gene' is the id currently used to generate identifiers urls but it should actually be 'ncbi'

src/client/features/paint/pathway-node-metadata.js
```diff
@@ -1,81 +1,120 @@
 const React = require('react');
 const h = require('react-hyperscript');
+const _ = require('lodash');
+
+const { generateIdentifiersUrl } = require('../../common/external-service-data');
+const { ServerAPI } = require('../../services');
+
+const DEFAULT_NUM_NAMES = 3;
+const SUPPORTED_DB_LINKS = {
```
d2fong: This should be a part of the functionality in 'external-service-data'

src/client/common/external-service-data.js
```diff
@@ -1,28 +1,20 @@
...
-  {database:'DrugBank', url:'https://www.drugbank.ca/', search:''},
-  {database:'mirtarBase', url:'http://identifiers.org/mirtarbase/', search:''},
-  {database:'NetPath', url:'http://www.netpath.org/', search:'molecule?molecule_id='},
-  {database:'Panther', url:'http://pantherdb.org/', search:'genes/geneList.do?searchType=basic&fieldName=all&organism=all&listType=1&fieldValue='},
-  {database:'PID', url:null},
-  {database:'PhosphoSitePlus', url:null},
+const { IDENTIFIERS_URL } = require('../../config');
```
d2fong: this file should also be renamed to something related to 'identifiers', not the general 'external-services-data'

src/server/routes/pathways/generate-pathway-json/biopax-metadata/index.js
```diff
@@ -1,226 +1,145 @@
...
+  );
+
+  let standardName = entryStdName || entRefStdName || '';
+  let displayName = entryDisplayName || entRefDisplayName || '';
+
+  return {
+    comments,
+    synonyms,
+    datasource,
+    type: type.includes(':') ? type.split(':')[1] : type,
```
d2fong: - maybe add a comment here explaining the type processing e.g. 'bp:Protein' -> 'Protein'
- also TODO need to split camel case type e.g. 'BiochemicalReaction' -> 'Biochemical Reaction'

Issue #1121, opened 2018-10-25
gconvert validation errors


d2fong:
Search on my local dev and appsbeta.pathwaycommons.org is not working.  

I think it is related to the gconvert validation service.  
apps.pathwaycommons.org still works. 

Here is the error I get:

```js

error:   A cache failed to be filled with args
error:   { '0': [ 'P04637' ], '1': { target: 'NCBIGene' } }
error:    TypeError: Cannot read property 'substring' of undefined
    at cleanUpEntrez (/Users/dylan/work/app-ui/src/server/external-services/gprofiler/clean-up-entrez.js:5:20)
    at _.forEach.info (/Users/dylan/work/app-ui/src/server/external-services/gprofiler/gconvert/index.js:75:20)
    at arrayEach (/Users/dylan/work/app-ui/node_modules/lodash/lodash.js:508:11)
    at Function.forEach (/Users/dylan/work/app-ui/node_modules/lodash/lodash.js:9334:14)
    at bodyHandler (/Users/dylan/work/app-ui/src/server/external-services/gprofiler/gconvert/index.js:72:5)
    at <anonymous>
    at process._tickCallback (internal/process/next_tick.js:189:7)
```

d2fong:
It seems like gconvert is returning a bad response:

Here is what is being returned from gconvert
```
[ '<html><body><h1>504 Gateway Time-out</h1>' ]
[ 'The server didn\'t respond in time.' ]
[ '</body></html>' ]
[ '<html><body><h1>504 Gateway Time-out</h1>' ]
[ 'The server didn\'t respond in time.' ]
[ '</body></html>' ]
```

d2fong:
Postman gives the same error:

<html>
    <body>
        <h1>504 Gateway Time-out</h1>
The server didn't respond in time.
    </body>
</html>

d2fong:
But it is strange that apps.pathwaycommons is not down also, since I think it uses the same gprofiler service.  Maybe the requests are cached.

maxkfranz:
TODO cache only single gene queries for validation

maxkfranz:
> api so that it returns 500 status and string as body rather than JSON

I think this is done given the changes we worked on yesterday on your machine.

PR #1122, opened 2018-10-25
Added util called safeFetch

jvwong:
Refs #1121 
- wrapper for global fetch instance to throw on response.ok is false


maxkfranz:
Note: The change we made makes it such that `global.fetch` is defined on the server side.

Let's be careful in future to make sure to always use the `util.fetch` so that the request fail when expected.

PR #1124, opened 2018-10-26
Iss1121 validator fixes

jvwong:
refs #1121 

- fetch wrapper (@maxkfranz) that handles HTTP status codes other than 200 
- enrichment services 
  - validation
    - test data processing; exclude any live fetch from g:Profiler 
    - `targetDb` input parameter must be MIRIAM preferred name
    - new `InvalidParamError` subclass 
    - added timeout for g:Convert call via config constant `GPROFILER_FETCH_TIMEOUT` (browser fetch has no timeout?)  
- pathway-commons services
  - let search algorithm bypass failures in validatorGconvert


jvwong:
I want to break this into smaller PRs...

Issue #1125, opened 2018-10-29
Tests: Remove live API calls to third parties (Enrichment, Summary)


jvwong:
Certain services fetch from third-party instances, but we should instead be testing the local  processing assuming a successful call.

This applies to both /enrichment and /summary code.

I'll leave the pc/interaction test calls and update mocha to obtain a global fetch instance.

maxkfranz:
Approaches:

(1) Create a mock `fetch()` function that just loads a predetermined file per test.  Since we're using `global.fetch()` on the server this should be straightforward.

(2) Refactor, where necessary, the services into separate raw fetch functions and processing functions.  Only run the processing functions in the tests.

jvwong:
I'm going to do (2) for enrichment services.

PR #1127, opened 2018-10-29
Iss1125 - Tests: Remove third-party http requests, Refactoring validator

jvwong:
- references #1125 
  - Validation
    - point URL to current version of g:Profiler
    - dropped 'organism' from API
    - Refactored and expose data processing code
    - Tests do not involve live g:Convert calls  
  - Tests
    - added `mocha.opts` pointing to global fetch instance
    - stubbed out 
      - `/api/summary/entity` 
      - `/api/enrichment/analysis`


src/config.js
```diff
@@ -21,7 +21,14 @@ let defaults = {
   DB_CERT: undefined,
   // factoid specific urls
   FACTOID_URL: 'http://unstable.factoid.baderlab.org/',
-  BIOPAX_CONVERTERS_URL: 'http://biopax.baderlab.org/'
+  BIOPAX_CONVERTERS_URL: 'http://biopax.baderlab.org/',
+  DATASOURCE_NAMES :{// !!!temporary , will be updated as part of 'identifiers url generation module' https://github.com/PathwayCommons/app-ui/issues/1116
```
maxkfranz: You can't have embedded objects with `config`.  An env var can't specify an object -- only a string value.

maxkfranz: Why not just use a consistent value instead?

E.g. just `HGNC_SYMBOL` -- never `HGNCSYMBOL`.

src/server/routes/enrichment/index.js
```diff
@@ -211,281 +211,21 @@ enrichmentRouter.post('/visualization', (req, res) => {
...
- *           - tcacao
- *           - tpratense
- *           - turartu
- *           - vvinifera
- *           - zmays
+ *           - "ENSEMBL"
+ *           - "HGNC"
+ *           - "HGNCSYMBOL"
+ *           - "UNIPROT"
+ *           - "NCBIGENE"
```
d2fong: we don't need to document species anymore?

test/server/enrichment/validation/validation-test.js
```diff
@@ -1,71 +1,137 @@
...
-		]
-	},
-	"alias": {
-    "TP53": "HGNC:11998",
-    "ATM": "HGNC:795"
-  }
+const { DATASOURCE_NAMES } = require('../../../../src/config');
+const { getForm, mapParams, gConvertResponseHandler } = require ('../../../../src/server/external-services/gprofiler/gconvert');
+const InvalidParamError = require('../../../../src/server/errors/invalid-param');
+const GPROFILER_DATASOURCE_NAMES = {
```
maxkfranz: This should be exposed in the src somehow rather than being duplicated here.

src/server/routes/enrichment/index.js
maxkfranz: I think we're only supporting human, since PC is generally human-only.

maxkfranz:
The above comments may make sense to resolve in a different ID-related PR.

src/server/routes/enrichment/index.js
jvwong: On line 214 for `query` I say your input has to be human. 
Also, g:Convert is tricky as it subsets the `target` database field based on species, so I just removed it.

Issue #1128, opened 2018-10-29
Server-side module that wraps pc xref-to-uri mapper (/pc2/miriam/uri) 


jvwong:
## Background

There are instances in the app where we provide hyperlinks for an entity. These hyperlinks are intended to resolve to a human-readable web page concerning a 'record' (e.g. P04637) within a 'collection' ('UniProt Knowledgebase') from a 'resource' (e.g. 'Universal Protein Resource using Persistent URL system'). There, users can learn more about the entity and link to resources therein.

The Pathways node tooltip data derives from BioPAX Xrefs. However, it is more useful to supply clients with  label - url pairs, rather than db - id pairs. This could be aided by using the [pc metadata web service](http://www.pathwaycommons.org/pc2/swagger-ui.html#!/metadata45controller/identifierOrgUriUsingGET) to map Xref db, id pairs  to  a resolvable url.    

## Description of new feature

A server-side module that wraps the pc metadata web service. Takes in (xref) db and id and returns url.

## Motivation for new feature

Towards a more robust approach to map pathway xrefs to label, url pairs that can be sent to clients. This is used in pathway metadata tooltip.

IgorRodchenkov:
@jvwong you may now try and use the web service described at http://biopax.baderlab.org/docs/index.html
instead of PC2/miriam/... The new one is provided with the BioPAX Validator app and can check multiple xrefs via POST request: just submit only 'db' and 'id' object fields in the query data (json array); in retun, you get the list of checked and updated "xref" objects, in the same order, where you can find additional fields: 'uri', 'namespace', 'preferredDb' for the valid (dbOk=true, idOk=true) ones.

jvwong:
COol. I'll open new ticket for swapping out xref2uri implementation.

jvwong:
@IgorRodchenkov what's the URL?

PR #1130, opened 2018-10-29
Adding xref2uri module to pc external services

jvwong:
refs #1128 
xref2uri module that wraps [pc metadata service](http://www.pathwaycommons.org/pc2/swagger-ui.html#!/metadata45controller/identifierOrgUriUsingGET) for mapping Xrefs db, id to url


jvwong:
CI failure addressed in #1127

Issue #1129, opened 2018-10-29
Pathway: Produce and consume label - url pairs derived from Xrefs


jvwong:
## Background

This issue depends upon #1128.

## Description of new feature

Pathway service should provide node metadata with label - url pairs derived from Xrefs. Client should consume and display this metadata. 

## Motivation for new feature

More robust means to provide and display pathway node metadata database links derived from Xrefs.

Issue #704, opened 2018-04-23
View node tooltip: to show


IgorRodchenkov:
@jvwong @d2fong @logmosier @gbader 

Here is what a UniProt person suggested:

> I searched with  "Leucine" and then clicked on  "Valine, leucine and isoleucine biosynthesis (KEGG pathway)", then clicked on the  IARS icon in the diagram, and then on  "links UniProt" and get Q7L4K8_HUMAN. This doesn't seem to be the best entry to show - it is a TrEMBL entry (unreviewed) and not a reviewed Swiss-Prot entry, and it also is a fragment. SYIC_HUMAN would be a much better choice!.. I can see that you actually have the full list of identifiers once I click on "More", but it would be good to have the Swiss-Prot entry stand out, and select it as the representative in the short view where you are only showing one link."

I know, right now, it's probably uneasy to tell a primary ID from others there (using the BioPAX, SBGN, or JS data from PC)... Though, I might be wrong as I recall we had some javascript code/function that checks IDs and perhaps maps to primary, no? If not, I could somehow label primary UniProt ID xrefs in Pathway Commons BioPAX and in the [physical_entities.json](http://www.pathwaycommons.org/archives/PC2/v9/physical_entities.json.gz) file in future PC db/webservice releases.

logmosier:
Ya, I should have quoted what I was referring to 

> Though, I might be wrong as I recall we had some javascript code/function that checks IDs and perhaps maps to primary, no?

PR #1134, opened 2018-10-30
Iss1129 Produce and consume label - url pairs derived from Xrefs

jvwong:
refs #1129 

- pathway service uses new `xref2uri` wrapper to fetch urls for entity ids
- pathway tooltip uses this `xrefLink` data 
- tests target pathway data processing -  not live fetch


src/server/routes/pathways/generate-pathway-json/biopax-metadata/index.js
```diff
@@ -35,18 +36,30 @@ let extractBiopaxMetadata = biopaxJsonEntry => {
...
+    xrefLinks
   };
 };
 
+// helper for converting xrefs to map of uris grouped by db
+const xref2link = ( db, id, xrefLinks ) => {
+  return xref2Uri( db, id )
+    .then( uri => {
+      if( xrefLinks[ db ] != null ){
+        xrefLinks[ db ] = xrefLinks[ db ].concat( uri );
```
jvwong: This relates to #704. I am just sticking all these xrefs in here and letting the client decide. Right now the client just picks the first one for each db...

d2fong: I am concerned that this would mean that we would be making potentially hundreds of requests to generate this pathway json now.

jvwong: Fair enough! Alternatives:
  - use the external-service-data mapping within `xref2Uri` rather than live service
  - get @IgorRodchenkov to spin up an endpoint to take in multi-pairs of xrefs

maxkfranz: I'd prefer the second option.  The convertor is nice for consistency but it needs to be fast.

IgorRodchenkov:
I could also generate a mapping CSV file: db, id, uri 
for all the xrefs in PC model.
Let’s talk requirements and design tomorrow.

IgorRodchenkov:
Well, in fact, a mapping file from db names to identifiers.org namespaces would suffice (if we don’t validate xref IDs in this app, and I think we should not).
E.g., ‘uniprot’ maps to ‘uniprot’, ‘kegg compound’ - ‘kegg.compound’, ‘uniprot isoform’ - ‘uniprot.isoform’ (unfortunately, there are cases where simple replacing of a space with period and turning to lower case would be wrong)

jvwong:
Yeah I think you're right - mapping file is enough. A service would get ugly.

IgorRodchenkov:
> Yeah I think you're right - mapping file is enough. A service would get ugly.

Why ugly? Unnecessary, yes, unless we really wanted to detect misspelled db names, validate IDs, etc.

jvwong:
I was thinking about the IDs but I guess they are pre-validated. So basically just need a way to do (preferred) name to namespace mapping.

src/server/external-services/pathway-commons.js
```diff
@@ -121,23 +125,70 @@ const sifGraph = async ( queryObj ) => {
...
+  const url = config.PC_URL + 'pc2/miriam/uri/' + constructQueryPath( name, localId ) ;
+  return fetch( url , { method: 'GET', headers: { 'Accept': 'text/plain' } })
+    .then( res => res.text() )
+    .then( handleEntityUriResponse );
+};
+
+// Custom cache since 1. want to hash off only one param 2. cases I don't wish to delete on error
+// I hacked the cache previously to hash off a single param index, but wasn't sure...
+const rawGetEntityUriBase = ( name, localId ) => {
+  let hash = hasher.hash( name );
```
jvwong: This is the only difference between the existing cache

IgorRodchenkov:
@jvwong  
Aha, I forgot that there is identifiers.org web service where you can get all the info about the Identifiers.org/MIRIAM data collections at once:
`https://identifiers.org/restws`
- 'name', 'synonyms', 'url' (alternatively - 'prefix') are fields we want :)

IgorRodchenkov:
@jvwong 
e.g.
`https://identifiers.org/rest/collections/`
however, some names map to many collections, such as 'kegg' maps to kegg.compound, kegg.genes, etc. (unlike it's in pc2/miriam/ where we removed ambiguous synonyms, a.f.a.i.k.):
`https://identifiers.org/rest/collections/name/kegg`

jvwong:
Yeah I tried to use identifiers.org services - I would basically have to do the work that the pc2 service does. For instance, with  all to https://identifiers.org/rest/collections/name/mi I gotta look at `name`, then maybe try `synonyms`, then check against their regex `pattern` .... Plus now I'm depending on identifiers.org being up etc. I just deleted it and put the current version up.

src/server/external-services/pathway-commons.js
maxkfranz: If the name uniquely identifies the query, then hashing isn't necessary.  Hashing is useful when you have a large query tuple or if the strings would cost a lot to store.  In this case, it may be better to just use the name as the key so that everything is human-readable -- that's good for debugging.

src/server/external-services/pathway-commons.js
```diff
@@ -121,23 +125,70 @@ const sifGraph = async ( queryObj ) => {
...
+  if( pcCache.has( hash ) ){
+    return pcCache.get( hash );
+  } else {
+    let res = fetchEntityUriBase( name, localId );
+    pcCache.set( hash, res );
+    res.catch( err => {
+      pcCache.del( hash );
+      logger.error(`Failed to fill cache with ${name} and ${localId} - ${err}`);
+    });
```
maxkfranz: If we don't already have a util function to cache a fetch call, then we should add one later to avoid repetition.  Factoid, for example, uses a util memoize function that accepts a custom cache per `memoize()` call.  Then you just wrap the fetch function.

d2fong:
Is there any problem with modifying the service to take multiple pairs of xrefs?

maxkfranz:
According to a discussion with Jeff yesterday: We shouldn't need to make a query per entity anymore.  The server should just get the list of namespaces from MIRIAM and then use identifiers.org links.

jvwong:
A little more info: For each xref, a request is made to `xref2uri` for the URI supplying a db, id pair. If that db is not in the cache (or memoized in the future), it goes out to the `/pc2/miriam/uri` service and gets it. 

In practice, after you fetch a couple of pathways (i.e. you see UniProt, HGNC, ...)  it basically never has to fetch.

src/server/external-services/pathway-commons.js
```diff
@@ -121,23 +125,70 @@ const sifGraph = async ( queryObj ) => {
...
+/* fetchEntityUriBase
+ * Light wrapper around the pc2 service to get the uri given a collection name and local ID for entity
  * http://www.pathwaycommons.org/pc2/swagger-ui.html#!/metadata45controller/identifierOrgUriUsingGET
+ * NB: pc2 service returns 200 and empty body if collection name and/or local ID are unrecognized.
+ *   If the local ID is empty, throws a 404
+ * @return { object } a node URL object
+ */
+const fetchEntityUriBase = ( name, localId ) => {
+  //console.log( `fetchEntityUriBase with collection = ${name}; ID = ${localId}` );
+  const url = config.PC_URL + 'pc2/miriam/uri/' + constructQueryPath( name, localId ) ;
```
d2fong: @IgorRodchenkov  @jvwong does this return a result if I send the name 'uniprot knowledgebase' because that is one of the names that uniprot appears under in the jsonld biopax.


See: 

https://github.com/PathwayCommons/app-ui/blob/development/test/server/pathways/input.biopax.json#L564

d2fong: Nevermind, it does
http://www.pathwaycommons.org/pc2/miriam/uri/uniprot%20knowledgebase/P27695

I just got confused because Igor's comment here said that 'uniprot' maps to 'uniprot':

```
E.g., ‘uniprot’ maps to ‘uniprot’, ‘kegg compound’ - ‘kegg.compound’, ‘uniprot isoform’ - ‘uniprot.isoform’ (unfortunately, there are cases where simple replacing of a space with period and turning to lower case would be wrong)
```

src/server/external-services/pathway-commons.js
```diff
@@ -121,23 +125,70 @@ const sifGraph = async ( queryObj ) => {
...
+
+/* fetchEntityUriBase
+ * Light wrapper around the pc2 service to get the uri given a collection name and local ID for entity
  * http://www.pathwaycommons.org/pc2/swagger-ui.html#!/metadata45controller/identifierOrgUriUsingGET
+ * NB: pc2 service returns 200 and empty body if collection name and/or local ID are unrecognized.
+ *   If the local ID is empty, throws a 404
+ * @return { object } a node URL object
+ */
+const fetchEntityUriBase = ( name, localId ) => {
+  //console.log( `fetchEntityUriBase with collection = ${name}; ID = ${localId}` );
```
d2fong: delete log statements

src/server/external-services/pathway-commons.js
```diff
@@ -121,23 +125,70 @@ const sifGraph = async ( queryObj ) => {
   });
 };
 
-/* xref2Uri
- * Light wrapper around pc metadata service for mapping Xrefs db, id to url
+const handleEntityUriResponse = text => {
+  if( !text ) return ''; // Go ahead so we can cache, but gotta throw to inform clients
```
d2fong: So this is caching empty responses for invalid inputs?

src/server/external-services/pathway-commons.js
jvwong: Indeed a bunch of possible collection names (e.g. UniProt Knowledgebase) map to the same namespace (e.g. uniprot) : MIRIAM 'preferred', MIRIAM 'synonym', PSI-MI controlled vocabulary 'database citation'. What rules they use for namespace is not clear.  E.g. 'Enzyme Nomenclature' maps to 'ec-code'...

src/server/routes/pathways/generate-pathway-json/biopax-metadata/index.js
```diff
@@ -35,18 +36,31 @@ let extractBiopaxMetadata = biopaxJsonEntry => {
     type: type.includes(':') ? type.split(':')[1] : type,
     standardName,
     displayName,
-    databaseIds
+    xrefLinks
   };
 };
 
+// helper for converting xrefs to map of uris grouped by db
+const xref2link = ( db, id, xrefLinks ) => {
```
d2fong: Generally I think that it is not a good idea to have functions like this produce side effects.  In my opinion it makes the code harder to understand.  
As a general guideline going forward I think it is good to keep functions that:
1. Are generally independent reusable pieces of functionality.
2. Have clear inputs and outputs.
3. Have Minimal side effects.
 
If I was reading this function for the first time I would be confused and ask:
1.  Why does this function update a `xrefLinks` object?
2.  Where is this function being used?
3.  What is the context of this function?  

The answer to my knowledge is that it is being used only in one place and the reason xrefLinks is being updated is because it is preprocessing all xrefs so that they can be easily accessed via each element id.  I think this information is obscured in the `xref2Link` function and therefore I think it is better if this code is just inlined into the`biopaxText2ElementMap` function.

i.e.

```
// process all xrefId elements and get a uri for each of them
xrefIds.fiilter( xrefId => xrefId != null ).forEach( xrefId => {
  let { k, v } = xRefMap.get( xrefId );
  let xrefUri = await xref2Uri( k, v );
  if( dbIds[k] != null ){
    dbIds[k] = dbIds[k].concat( v );
  } else {
    dbIds[k] = [v];
  }
});
```

test/server/pathways/pathways-network-generation.js
```diff
@@ -1,16 +1,27 @@
...
-const { getPathwayJson } = require('../../../src/server/routes/pathways/generate-pathway-json');
+const { expect } = require('chai');
+const fs = require('fs');
+const path = require('path');
+const { getBiopaxMetadata } = require('../../../src/server/routes/pathways/generate-pathway-json/biopax-metadata');
+const sampleCyjsonData = require('./sample-cyjson-data');
+const sampleMetadataOut = require('./sample-metadata-out.json');
 
+describe('Pathways network generation', function(){
+  beforeEach(function() {
```
d2fong: whats the purpose of this beforeEach?

IgorRodchenkov:
> Is there any problem with modifying the service to take multiple pairs of xrefs?

@d2fong @jvwong @maxkfranz 

I am making the change right now, but it is not going to be /pc2/miriam (that would require a new PC release or beta server). I think it belongs to the biopax validator, and the endpoint will be, e.g., biopax.biopax.org/validator/recommend/xref/ (accept the list of db names or db,id pairs to return a json doc mapping input/all to preferred db names, namespaces, URIs) or alike (I am even thinking of using GraphQL...). But, I could quickly make a simpler version of it just like current one but with JSON input/output. (See BioPAX/validator#4)

Tell me **how you'd prefer** submit a list of "xrefs" (db,id pairs) to the service?

IgorRodchenkov:
> Yeah I tried to use identifiers.org services - I would basically have to do the work that the pc2 service does. For instance, with c all to https://identifiers.org/rest/collections/name/mi I gotta look at `name`, then maybe try `synonyms`, then check against their regex `pattern` .... Plus now I'm depending on identifiers.org being up etc. I just deleted it and put the current version up.

I mean, we could parse the https://identifiers.org/rest/collections/ result once to make the name/namespace-to-namespace map once and use it to build all URIs (and skip for all invalid names)...

src/server/external-services/pathway-commons.js
jvwong: I can't reproduce the case for justifying this. Removing...

jvwong: Now I found it again: http://www.pathwaycommons.org/pc2/get?uri=http%3A%2F%2Fidentifiers.org%2Freactome%2FR-HSA-69541

This pathway contains 3 xrefs with a similar `db` field:
```xml
	<bp:RelationshipXref rdf:ID="RelationshipXref_reactomehttp___www_reactome_org_biopax_64_48887_RelationshipXref1782">
		<bp:relationshipType rdf:resource="#RelationshipTypeVocabulary_ac7de6f2f302971b64781fc96cc97c86" />
		<bp:comment rdf:datatype = "http://www.w3.org/2001/XMLSchema#string">REPLACED http://www.reactome.org/biopax/64/48887#RelationshipXref1782</bp:comment>
		<bp:id rdf:datatype = "http://www.w3.org/2001/XMLSchema#string">6804946</bp:id>
		<bp:db rdf:datatype = "http://www.w3.org/2001/XMLSchema#string">reactome database id release 64</bp:db>
	</bp:RelationshipXref>
```

I was caching this because I was concerned about how often these types of things make their way into BioPAX. Also I was  assuming this function wouldn't be exposed as an endpoint, so everything that comes in should be OK.

src/server/routes/pathways/generate-pathway-json/biopax-metadata/index.js
jvwong: Good point - the Promise.all calls were resulting in a bunch of cache misses for the same db as the same xrefs dbs were being sent out at the same time. Made serial and inlined..

test/server/pathways/pathways-network-generation.js
jvwong: none. removed

src/server/external-services/pathway-commons.js
IgorRodchenkov: simply ignore all the xrefs having db="reactome database id release 64" (or starts with "reactome database id release"), because these are internal and not useful xrefs, and there is always a standard stable xref with db="reactome" and id like "R-HSA-123456"...

d2fong: how does having a db field with 'reactome database id release 64' hit the empty case for this function?

jvwong: The fetch body.text() will be empty, but return HTTP 200.

d2fong: oh because reactome database id doesnt have a mapping?

jvwong: well I think what happening here is that Reactome is 'abusing' this Xref and basically putting comments rather than a valid db, id.

maxkfranz: If we get an empty response, we should consider throwing an error.

jvwong: Latest commit I let those empty responses throw up to caller.  Do not get cached.

PR #1135, opened 2018-10-31
add timeouts for fetch requests

d2fong:
- use bluebird timeout to prevent continuously waiting on requests when the corresponding external services may be down
- add a fetch timeout config value (default 5 seconds)
- tested by replacing the ncbi url with 'http://localhost:3000/api/test' which waits 10 seconds before sending a response, the request successfully timed out and logs the following error: 
`error:   TimeoutError in ncbi fetchByGeneIds: operation timed out` Not sure if there is a better way to test the behaviour.

- from my somewhat limited testing, the client doesn't hang anymore when the server takes more than 5 seconds to respond

### DIFF
--- a/src/server/routes/index.js
+++ b/src/server/routes/index.js
@@ -10,6 +10,9 @@ router.use('/api/pathways', require('./pathways'));
 router.use('/api/interactions', require('./interactions'));
 router.use('/api/enrichment', require('./enrichment'));
 router.use('/api/factoids', require('./factoids'));
+router.get('/api/test/', (req, res) => { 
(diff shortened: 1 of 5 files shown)


src/server/routes/index.js
```diff
@@ -10,6 +10,9 @@ router.use('/api/pathways', require('./pathways'));
 router.use('/api/interactions', require('./interactions'));
 router.use('/api/enrichment', require('./enrichment'));
 router.use('/api/factoids', require('./factoids'));
+router.get('/api/test/', (req, res) => {
```
maxkfranz: Are we keeping this for the mocha tests, or should we remove it?

Issue #1139, opened 2018-11-02
Use p-memoize() with LRU caches for fetches


maxkfranz:
## Description of new feature

Use p-memoize to cache fetch() calls: https://github.com/sindresorhus/p-memoize

## Motivation for new feature

- Avoids duplicate code for using LRU caches with fetch().
- Already handles promise rejection.
- Can just pass in LRUCache to memoize()

jvwong:
> cache
> - Type: Object
>  - Default: new Map()
>  - Use a different cache storage. Must implement the following methods: .has(key), .get(key), .set(key, value), .delete(key), and optionally .clear(). You could for example use a WeakMap instead or quick-lru for a LRU cache.

In the app we use lru-cache which defines a `del(key)` rather than `delete(key)` - use this [quick-lru](https://www.npmjs.com/package/quick-lru) instead?

maxkfranz:
quick-lru should be fine for what we're doing.

maxkfranz:
May as well ditch lru-cache then

jvwong:
Small update required to `src/server/external-services/pathway-commons.js` to accomodate quick-lru pacakge API.

Issue #1144, opened 2018-11-07
Validation: Sort query before caching


jvwong:
The goal is to cache validation queries in a more efficient manner. As a first step, sort the query token array alphabetically before it hits the cache. Currently, something like ['TP53', 'MDM2'] is not the same as ['MDM2', 'TP53'].

maxkfranz:
options.sort: true in node-object-hash

https://www.npmjs.com/package/node-object-hash#constructor-requirenode-object-hashoptions

jvwong:
The docs seem to say options.sort = true by default?

>Parameters:
> options:object - object with hasher config options
options.coerce:boolean - if true performs type coercion (default: true); e.g. hash(true) == hash('1') == hash(1), hash(false) == hash('0') == hash(0)
options.sort:boolean - if true performs sorting on objects, arrays, etc. (default: true);
options.alg:string - sets default hash algorithm (default: 'sha256'); can be overridden in hash method;
options.enc:string - sets default hash encoding (default: 'hex'); can be overridden in hash method;

maxkfranz:
Since the option is important, I'd be inclined to make the option explicit in the constructor in case we upgrade the package in future and they change the defaults.  (Some package authors consider an undefined option to mean "I don't care about this option" versus "I'm OK with this specific default value".)

It's not strictly necessary though.

jvwong:
Will declare in options explicitly.

PR #1145, opened 2018-11-07
Use p-memoize with quick-lru; defaults

jvwong:
refs #1139 - swap out p-memoize for cache implementation
- used quick-lru to satisfy p-memoize (.delete function)


maxkfranz:
This looks good to merge to me

maxkfranz:
Could potentially include  this in the PR since it would just be a tiny change: Validation: Sort query before caching #1144

Issue #1137, opened 2018-11-02
Handle (safe) fetch TimeoutErrors


jvwong:
This involves decisions about how to handle errors from PC and third-party services/wrappers.

PC-side:
  - search
  - validator
  - interactions

Third-party services:
  - g:Profiler
  - genenames.org, NCBI, UniProt etc

jvwong:
Part 1: Approach for pathway search list results 
  - Obtaining a pathway list from a query depends on a. PC web service and b. app-ui [validation/g:Convert](https://github.com/PathwayCommons/app-ui/blob/266cca292f65b042ae99469374e100182c7fa1cc/src/server/external-services/pathway-commons.js#L26-L42)
  - Let the validation step fail,  providing a reasonable fallback

jvwong:
Part 2: Approach for entity summary box
  - Obtaining the info box depends on validator 
  - Swallow any Errors, thus enabling app to display everything else (pathway list)

jvwong:
You're right, Part 2 is swallowed in the `ServerAPI.entitySummaryQuery()`. Part 1 is still necessary - app hangs if anything is thrown in validator/g:Convert.

PR #1151, opened 2018-11-08
Search timeout handling

d2fong:
- catch errors in the gconvert entity extraction and return an empty array if if fails
- make the code quality in pathway-commons service more consistent with the rest of the codebase

refs #1137

### DIFF
--- a/src/server/external-services/pathway-commons.js
+++ b/src/server/external-services/pathway-commons.js
@@ -17,50 +17,60 @@ const fetchOptions = {
   }
 };
 
-const _sanitize = (s) => {
+//Pathway Commons HTTP GET request; options.cmd = 'pc2/get', 'pc2/search', 'pc2/traverse', 'pc2/graph', etc.
+let query = opts => {
+  let queryOpts = _.assign( { user: 'app-ui', cmd: 'pc2/get' }, opts);
+  let { cmd } = queryOpts;
+  let url = config.PC_URL + cmd + '?' + qs.stringify( queryOpts );
+
+  return fetch(url, fetchOptions)
+    .then(res => ( cmd === 'pc2/get' || cmd === 'pc2/graph' ? res.text() : res.json() ) )
+    .catch((e) => {
+      logger.error('query ' + queryOpts + ' failed - ' + e);
+      return null;
+    });
+};
+
+let sanitize = s => {
   // Escape (with '\'), to treat them literally, symbols, such as '*', ':', or space,
   // which otherwise play special roles in a Lucene query string.
   return s.replace(/([!*+\-&|()[\]{}^~?:/\\"\s])/g, '\\$1');
 };
 
-const _processPhrase = (phrase) => {
-  return validatorGconvert(phrase.split(' '),{}).then(result => {
+
+// recognize biological entities from an input string
+let extractEntityIds = inputString => {
+  let tokens = inputString.split(' ');
+
+  return validatorGconvert( tokens ).then( result => {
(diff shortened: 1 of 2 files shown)


src/server/external-services/pathway-commons.js
```diff
@@ -17,74 +17,87 @@ const fetchOptions = {
...
+
       return recognized ? ( 'xrefid:' + sanitized ) : ( 'name:' + '*' + sanitized + '*' );
     });
+
     return entities.concat(otherIds);
+  })
+  .catch( e => {
+    logger.error('unable to get response from gconvert with the following inputstring: ' + inputString);
+    logger.error(e);
+    return [];
```
d2fong: Not sure if it is a good idea to return stuff from a catch.

src/server/external-services/pathway-commons.js
```diff
@@ -17,74 +17,87 @@ const fetchOptions = {
...
+let search = async opts => {
+  let { gt:minSize = 0, lt:maxSize = 250, q } = opts;
+  let queryStrings = await generateSearchQueries( q.trim() );
+
+  for( let queryString of queryStrings ) {
+    let queryOpts = _.assign( opts, { cmd: 'pc2/search', q: queryString } );
+    let searchResult = await query( queryOpts );
+    let searchResults = _.get( searchResult, 'searchHit', []);
+
+    if( searchResults.length > 0 ){
```
d2fong: This can be simplified by filtering directly outside of the if statement

src/server/external-services/pathway-commons.js
```diff
@@ -17,74 +17,87 @@ const fetchOptions = {
   }
 };
 
-const _sanitize = (s) => {
+//Pathway Commons HTTP GET request; options.cmd = 'pc2/get', 'pc2/search', 'pc2/traverse', 'pc2/graph', etc.
+let query = opts => {
+  let queryOpts = _.assign( { user: 'app-ui', cmd: 'pc2/get' }, opts);
```
d2fong: Might make sense to add these pc user opts to the top of the scope in this module.

src/server/external-services/pathway-commons.js
jvwong: If validator fails, I was thinking to return, for each item in `tokens`:
 `'name:' + '*' + sanitized + '*'`

d2fong: ok

src/server/external-services/pathway-commons.js
```diff
@@ -17,109 +17,108 @@ const fetchOptions = {
...
-      });
-      if (filteredResults.length > 0) {
-        return filteredResults;
-      }
+let search = async opts => {
+  let { gt:minSize = 0, lt:maxSize = 250, q } = opts;
+  let queryStrings = await generateSearchQueries( q.trim() );
+
+  for( let queryString of queryStrings ) {
+    let queryOpts = _.assign( opts, { cmd: 'pc2/search', q: queryString } );
```
jvwong: I think @IgorRodchenkov mentioned dropping `gt` and `lt` from `queryOpts` since PC WS doesn't use them.

src/server/external-services/pathway-commons.js
maxkfranz: It may be better to recover from the error closer to the UI, either in the UI directly (ideal) or in a higher level service that goes through all the pipeline steps.

d2fong: I think this is part of the search service pipeline.

PR #1155, opened 2018-11-13
Direct PC WS search; limiting query to 250 chars

jvwong:
This PR arises after experience with TimeOut errors with queries. We have decided to simplify the pathway search to basically forward queries to the PC Web Services. 

Have also identified users dumping extremely long gene lists via logs. We have restricted the character limit (250) to accommodate reasonable use case.


src/server/external-services/pathway-commons.js
```diff
@@ -31,78 +30,18 @@ let query = opts => {
...
-    '(' + entities.join(' OR ') + ')',
-    inputString //"as is" (won't additionally escape Lucene query syntax, spaces, etc.)
-  ];
-};
-
 // A fine-tuned PC search to improve relevance of full-text search and filter out unwanted hits.
 // The argument (query object) has the following fields:
 //  - q: user input - search query string
 //  - type: BioPAX type to match/filter by
 //  - lt: max graph size result returned
```
d2fong: update comments - no longer use lt, gt

Issue #1156, opened 2018-11-13
Unified 'query' service


jvwong:
Background: A goal is to have a 'smart search' in which information retrieved and displayed depends on the user input. For instance, entering a few biological IDs would return 'interactions' whereas for longer lists, the 'enrichment' results would be displayed.

Issue: Currently, search executes two parallel paths: One that populates the pathway hits and another for the summary. Making parallel data calls in the client makes it difficult to decide on the best data to retrieve and leads to repeated activity (validation). 

A single service endpoint should be created that coordinates:

  1. Identifying biological IDs in user query 
  2. Getting pathway hits from PC
  3. Getting entity / interactions / enrichment info 

![search_ 2 0 1](https://user-images.githubusercontent.com/4706307/48488092-f9dbd800-e7ed-11e8-904a-b9ec87082833.png)

Mentioned in  #1153.

PR #1159, opened 2018-11-15
Iss1156 Unified search

jvwong:
refs #1156 

This is a sketch for a new app service endpoint 'search'. It will coordinate 
  - pathway search: sanitize, send off to pc
  - info box data: tokenize, recognize genes, get entity summaries for interactions app
    - sets the stage for deciding what results based on number of genes (enrichment)

Right now takes in POST with `{"query": "TP53 is the best"`} and returns:

```js
{
  "card": {
    "entities": [
      {
        "namespace": "ncbigene",
        "displayName": "tumor protein p53",
        "localId": "7157",
        "description": "This gene encodes a tumor suppressor ...",
        "aliases": [
          ...
        ],
        "aliasIds": [
          "BCC7",
         ...
        ],
        "xrefLinks": [
          ...
        ]
      }
    ]
  },
  "pathways": [
    {
      "uri": "http://identifiers.org/reactome/R-HSA-3232118",
      "biopaxClass": "Pathway",
      "name": "SUMOylation of transcription factors",
      ...
    },
    ...
  ]
}

```

Then we can remove endpoints for `/api/summary/*`, `/api/pc/*` and just use internally.


jvwong:
Not a strong one, just trying to have a predictable output when we switch from say few genes (interactions) to many (enrichment). What do you think about this idea which I've seen NCBI use - return object with array of keys that declare the object keys available -
```js
{
  "types": ["pathways", "summaries", ...],
  "pathways": [...],
  "summaries": []
}
```

d2fong:
I guess thats okay, but whats the difference between that and just checking each key manually?

d2fong:
oh but I guess in your approach it means the client doesn't have to know that the 'pathways' and 'summaries' keys exist?

jvwong:
I guess the question is what's the best way to display info box/card content contingently (based on the genes recognized)?

src/server/routes/index.js
```diff
@@ -3,14 +3,14 @@ const router = express.Router();
 const pc = require('./pathway-commons-router');
 const summary = require('./summary');
 
-
+router.use('/api/search', require('./search'));
```
d2fong: search is currently organized under /api/pc if we want /api/search then we should just scrap /api/pc

jvwong: Eventually, I was thinking we should remove all other 'search' (`/pc`, `/summary`) endpoints and just use the modules directly within this one. Unless there is a reason to expose those?

jvwong:
That `interactions` key is probably not appropriate for enrichment, as the enrichment map doesn't really describe interactions. Maybe `networks` or ...?

d2fong:
Yeah that makes sense

src/server/routes/index.js
d2fong: Yeah that makes sense

jvwong:
That's not what I expected rebase to do.

d2fong:
did you want to revert it or no?

maxkfranz:
Rule of thumb : Rebase only on private branches.  Merge public branches.

E.g. rebase private `feature1` branch w.r.t. `development`:

```
git checkout feature1
git rebase development
```

E.g. merge `development` onto public `feature2` branch:


```
git checkout feature2
git merge development
```

jvwong:
I thought I was rebasing my fork's branch. How to revert?

maxkfranz:
It's simpler to just create a new branch off of dev and cherry pick the commits you want in the PR in the new branch.

d2fong:
I think this might work?
git checkout < git commit id from before>
git checkout push -f <your branch name>

jvwong:
I'm just cherry picking onto a new branch --> PR

PR #1166, opened 2018-11-16
Input box attribute typo for maxLength

jvwong:
I think this is a React related requirement for camelCasing.

image


maxkfranz:
The html attribute is `maxlength` but react wants to use `maxLength`

PR #1168, opened 2018-11-16
Iss1156 unified search

jvwong:
Replaces my botched #1159.

### DIFF
--- a/src/server/routes/search/search.js
+++ b/src/server/routes/search/search.js
@@ -27,20 +27,24 @@ const sanitize = ( rawQuery, maxLength = QUERY_MAX_CHARS ) => rawQuery.trim().su
 const tokenize = ( rawQuery, maxNum = QUERY_MAX_TOKENS ) => rawQuery.split(/,?\s+/).slice( 0, maxNum ); //  limit token size?
 
 // Take the entity summaries (summaries) and augment with xref corresponding to recommended name (name)
-// Only alter those summaries that have a mapping (alias).
-const fillInXref = async ( summaries, alias, name ) => {
-   const mappedIds = _.keys( alias );
-   for( const id of mappedIds ){
-    const eSummary = _.find( summaries, s => s.localId === id );
-    if ( eSummary ) await pc.xref2Uri( name, _.get( alias, id ) ).then( xref => eSummary.xrefLinks.push( xref ) );
-   }
+const fillInXref = async ( summaries, ncbiAlias, uniprotAlias, name ) => {
+  const tokensWithUniprot = _.keys( uniprotAlias );
+  for( const token of tokensWithUniprot ){
+    const ncbiGeneId = ncbiAlias[ token ];
+    const eSummary = _.find( summaries, s => s.localId === ncbiGeneId );
+    if ( eSummary ) {
+      // Use our internal service to grab the xref info
+      const xref = await pc.xref2Uri( name, _.get( uniprotAlias, token ) );
+      eSummary.xrefLinks.push( xref );
+    }
+  }
 };
 
 // Create an entity summary using NCBI, augmented with UniProt Xref
 const getNcbiSummary = async ( ncbiAlias, uniprotAlias ) => {
   const ncbiIds = _.values( ncbiAlias );
   const summaries = await entityFetch( ncbiIds, NS_NCBI_GENE );
-  await fillInXref( summaries, uniprotAlias, NS_UNIPROT );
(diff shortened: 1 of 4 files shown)


src/server/routes/search/search.js
```diff
@@ -0,0 +1,81 @@
...
+// Get the info related to gene-based interactions
+const geneInteraction = async entityIds => {
+  const summaries = await entityFetch( entityIds, NS_NCBI_GENE );
+  const sources = summaries.map( summary => hgncSymbolsFromXrefs( summary.xrefLinks ) );
+  return { sources, summaries };
+};
+
+// Return information about networks
+// Logic herein decides type of data (genes [, pathways])
+const searchNetworks = async query => {
```
d2fong: Search networks is ambiguous, we have multiple types of networks this function name should be more precise

jvwong: As per our discussion, let's leave the client to decide what to show or do. Here, service will just return any entity/gene summaries recognized.

src/server/routes/search/search.js
```diff
@@ -0,0 +1,106 @@
...
+  return symbol;
+};
+
+const sanitize = ( rawQuery, maxLength = QUERY_MAX_CHARS ) => rawQuery.trim().substring( 0, maxLength );
+const tokenize = ( rawQuery, maxNum = QUERY_MAX_TOKENS ) => rawQuery.split(/,?\s+/).slice( 0, maxNum ); //  limit token size?
+
+// Take the entity summaries (summaries) and augment with xref corresponding to recommended name (name)
+// Only alter those summaries that have a mapping (alias).
+const fillInXref = async ( summaries, alias, name ) => {
+   const mappedIds = _.keys( alias );
```
jvwong: Needs to be updated...

jvwong:
The `search` depends on two lower functions:

- `searchGenes` has a catch that returns empty array
- `searchPathways`  doesn't have any error handling

What would you recommend?

jvwong:
Follow up: `searchPathways` calls `pc.search` which swallows any fetch errors and returns `null`.

d2fong:
I am trying to say that if one of them succeeds and one fails, the whole request shouldn’t fail.  Does this currently happen?

d2fong:
I don’t know if returning an empty array for the gene summaries makes sense.  Doesn’t the successful result for gene summaries return an object?  For the pathway results it should return an empty array

jvwong:
Yes. 

If anything in the `geneSearch` fails, it fails gently and returns an empty array. In the `searchPathways`, if `pc.search` fetch fails  you get null back.

jvwong:
> I don’t know if returning an empty array for the gene summaries makes sense. Doesn’t the successful result for gene summaries return an object? 

Returns an array of 'info'  objects `[{ "summary": {...}, "query": "", "geneSymbol": "" },...,{...}]`
 
> For the pathway results it should return an empty array

Yeah I could change the pc.search function to return empty array instead of null, but maybe the client should handle errors from the backend (e.g. 500) ?

maxkfranz:
We shouldn't be changing which functions throw exceptions or reject promises just to be more convenient to a higher level service.  Each function should be as independent as possible, and we should use exceptions wherever possible for error cases to allow for more downstream flexibility.

The higher level service should catch the exceptions from lower level services and handle them in a way that makes sense for that particular higher level service.

d2fong:
Yeah I think you are right the client should handle that.  Maybe the geneSearch function should return null instead of an empty array?

jvwong:
To recap, `search` will swallow errors from `searchGenes` with null but, will pass along anything (error, data) from `searchPathways`.

jvwong:
> geneSearch() should probably do something like this:
>   - case: error => throw exception with meaningful message.

The thing that's holding me back is the higher level function, which afik is  'all-or-none':

```js
const search = async ( query ) => {
  return Promise.all([ searchGenes( query ), searchPathways( query ) ])
    .then( ([ genes, pathways ]) => ({ genes, pathways }) );
};
```

d2fong:
http://bluebirdjs.com/docs/api/promise.some.html . this function might work instead of all

maxkfranz:
```js
const searchGenesWithFallback = query => searchGenes(query).catch(() => []);
const searchPathwaysWithFallback = query => searchPathways(query).catch(() => []);

const search = async ( query ) => {
  return Promise.all([ searchGenesWithFallback( query ), searchPathwaysWithFallback( query ) ])
    .then( ([ genes, pathways ]) => ({ genes, pathways }) );
};
```

Issue #1153, opened 2018-11-12
Enrichment MVP


maxkfranz:
Let's discuss the list of things we need for a minimum viable product for the enrichment app.

(Feel free to edit these lists in the OP.)

Features:

- [ ] Remove sidebar
- [ ] Remove p-value slider.  Show p-value mini-visualisation within node tooltips.
- [ ] Fix layout.  It's way too spaced-out.
- [ ] Revise stylesheet so node sizes are more apparent.
- [ ] Revise the stylesheet so that edges are more consistent with the other apps.
- [ ] Remove info button.
- [ ] Make downloads a tooltip like other apps.  
- [ ] Support Cytoscape JSON download type
- [ ] Animated fit
- [ ] Improve clarity of GO linkout labels
- [ ] Each of 'Genes shared with entered list' and 'genes in pathway' should be linkouts
- [ ] Re-evaluate 'genes in pathway' display
- [ ] Clarify 'search pathway' button label
- [ ] The genes list box may not be necessary.  The input can be from the main search page.  Replace the search box with a title, like the other apps.
- [ ] Support for search results cards: SIF card, enrichment card, pathway card (pathway card optional)

Foundation:

- [ ] Clean up services (in progress)
- [ ] Clean up app code

@jvwong @d2fong

jvwong:
Sounds good. 

> The genes list box may not be necessary. 

Seems like it would be worth updating how the user query is plumbed. Right now, the user query  phrase is sent to the validation module independently (redundantly) by the 1. pathway search and 2. entity summary search services to identify IDs. What should happen is that the query should be sent to a validation module first, then forwarded to search and summary (and enrichment). In this way we could decided which cards to show based on number of genes that get recognized. Also more maintainable (see #1137).   

> Each of 'Genes shared with entered list' and 'genes in pathway' should be linkouts

Akin to our entity summary service, a pathway summary could collect all of this on the server (external services are GO and Reactome) and just let the client display the result in tooltip.

d2fong:
@maxkfranz I will work on the 'vis' portion.

maxkfranz:
OK, I'll take a look at tooltips

d2fong:
A few problems that I am noticing:
- nodes have really long labels
- the layout might look stretched out because there are potentially many disconnected components

d2fong:
layout options that seem to produce decent results:
```js
[...]

PR #1172, opened 2018-11-20
Enrichment style

d2fong:
refs #1153 

- use cola layout as it seems to produce better results for enrichment networks ( disconnected components)
- don't show text labels for the network until the user is zoomed past a specific font size
- edges and node styles are more consistent with the interactions app
- change the 'fit' button to be animated 
- add selection style for nodes and edges

screen shot 2018-11-20 at 6 09 10 pm
screen shot 2018-11-20 at 6 08 54 pm
screen shot 2018-11-20 at 6 08 39 pm


d2fong:
Not entirely sold on hiding node labels past a certain zoom.

PR #1186, opened 2018-11-27
Update nodemon so that we don't use flatmap-stream

maxkfranz:
https://github.com/dominictarr/event-stream/issues/116


maxkfranz:
I think we're at the point where no one should be committing directly to the master or dev branches of our projects.  Each change should be a PR that gets merged by someone other than the author so that there is some level of review.

PR #1203, opened 2018-12-04
Update to cytoscape@3.3.0

maxkfranz:
Not necessary right away, but it's nice to upgrade for testing etc.


maxkfranz:
We could hold off on merging this until after we make the next release.

Issue #1215, opened 2018-12-07
Search service: Keeps appending multiple instances of xrefLink for UniProt


jvwong:
## Current (buggy) behaviour

Each call to `search` service with same query appends additional xrefLink objects for uniprot to `summary`.

## Desired behaviour

Should only ever append uniprot xreflink once.

## Minimum steps to reproduce

POST to [/api/search](http://localhost:3000/api/search) with same `q` parameter in body:

```js
{ 
	"q": "MDM2 binds to TP53"	
}
```
Response: 
```js
{
  "genes": [
    {
      "query": "MDM2",
      "geneSymbol": "MDM2",
      "summary": {
        "namespace": "ncbigene",
       ...
        "xrefLinks": [
          {
            "namespace": "hgnc.symbol",
            "uri": "http://identifiers.org/hgnc.symbol/MDM2"
          },
          {
            "namespace": "genecards",
            "uri": "http://identifiers.org/genecards/MDM2"
          },
          {
            "namespace": "ncbigene",
            "uri": "http://identifiers.org/ncbigene/4193"
          },
          {
            "uri": "http://identifiers.org/uniprot/Q00987",
            "namespace": "uniprot"
          },
          {
            "uri": "http://identifiers.org/uniprot/Q00987",
            "namespace": "uniprot"
          },
          {
            "uri": "http://identifiers.org/uniprot/Q00987",
            "namespace": "uniprot"
          }
        ]
      }
    },
...
```

jvwong:
Problem is that the the `fillInXref` is called for every search request, regardless if lower-level function (entityFetch) is cached.

PR #1216, opened 2018-12-09
Simple check for existing uniprot xref

jvwong:
refs #1215 where duplicate xref being added to `xrefLinks`. Simple check before adding.


d2fong:
I think this change had unintended side effects.

Here are the genes I get when I search 'TP53 MDM2'
screen shot 2018-12-11 at 12 15 55 pm

Issue #1214, opened 2018-12-07
Enrichment: Auto-check g:GOSt for GMT file updates


jvwong:
Enhancement: Related to #1200

> How about using a periodic update (e.g. with node-cron)? We could update automatically on server startup and automatically once per week thereafter?

- Get an input stream, `f`, for the HTTP request to download the zip
- Pipe `f` to unzip: https://www.npmjs.com/package/unzip
- Once the stream has ended, pick out the human gmt file
- Cache the gmt file

_Originally posted by @maxkfranz in https://github.com/PathwayCommons/app-ui/issues/1200#issuecomment-444279629_

PR #1230, opened 2018-12-12
Iss1214 cron gmt update

jvwong:
refs #1214 

- Stubbed out node-cron on server
- Cron fires generic `update()` that calls enrichment's `updateEnrichment()` 
- Using [unzipper](https://www.npmjs.com/package/unzipper) to download g:Profiler pathway zip archive and extract file `hsapiens.pathways.NAME.gmt`
- Save to local file
- Using a small sample zip file for dev 

Looking for feedback on this. Open questions about this 'unzipper' library suitability. Also wondering if the data provided by `pathway-table` module will reflect the new file without reboot?


jvwong:
Let me reopen this one something went wrong.

PR #1231, opened 2018-12-12
Iss1214 cron update

jvwong:
refs #1214 

- Stubbed out node-cron on server
- Cron fires generic `update()` that calls enrichment's `updateEnrichment()` 
- Using [unzipper](https://www.npmjs.com/package/unzipper) to download g:Profiler pathway zip archive and extract file `hsapiens.pathways.NAME.gmt`
- Save to local file
- Using a small sample zip file for dev testing

Looking for feedback on this. Open questions about this 'unzipper' library suitability. Also wondering if the data provided by `pathway-table` module will reflect the new file without reboot?


src/server/index.js
```diff
@@ -9,8 +9,14 @@ const http = require('http');
 const stream = require('stream');
 const fs = require('fs');
 const Promise = require('bluebird');
+// const cron = require('node-cron');
 
 const config = require('../config');
+const update = require('./update');
```
d2fong: I would prefer that you be more specific in this module name.  It sounds too generic to the point of confusion.

If the goal of this module is group all tasks that we may need now and in the future that update periodically into a single module, then it needs a proper name.

src/server/index.js
```diff
@@ -9,8 +9,14 @@ const http = require('http');
 const stream = require('stream');
 const fs = require('fs');
 const Promise = require('bluebird');
+// const cron = require('node-cron');
```
d2fong: delete commented code

src/server/index.js
```diff
@@ -9,8 +9,14 @@ const http = require('http');
...
 
 const config = require('../config');
+const update = require('./update');
+
+// cron.schedule('10 * * * * *', () => {
```
d2fong: delete commented code

src/server/routes/enrichment/visualization/pathway-table.js
```diff
@@ -25,5 +29,11 @@ gmtPathwayData.split('\n').forEach( pathwayInfoLine => {
 
 pathwayInfoTable.delete('');
 
+/**
+ * handleFileUpdate
+ * @external file
+ * @see {@link https://www.npmjs.com/package/unzipper}
+ */
+const handleFileUpdate = file => file.stream().pipe( fs.createWriteStream( path.resolve( __dirname, file.path ) ) );
```
d2fong: Does it make sense for this function to be a part of the enrichment module?

src/server/index.js
jvwong: Something like `updateExternalDataFiles` or `refreshExternalDataFiles`?

src/server/routes/enrichment/visualization/pathway-table.js
jvwong: Alternative is `src/server/update.js` or `src/server/routes/enrichment/visualization/update-enrichment.js`?

src/server/index.js
jvwong: Checking if you mean delete or uncomment?

d2fong: I am assuming that if it is commented out then its not needed?   If its not needed it should be deleted.

src/server/index.js
d2fong: I guess so.  It depends how you see this module changing over time.  Do you think envision using more cron jobs in the future?  If so, then this should be named to imply that it is related to cron specific jobs.

maxkfranz: Maybe just `update-cron`

IgorRodchenkov:
@jvwong Jeff, are you still working on this PR and Dylan's comments (I simply tried to quickly resolve the merge conflicts...)?

jvwong:
Yes, i still want to complete this...

maxkfranz:
Let's review this in-person tomorrow.  I can help out with getting this finished.

jvwong:
@maxkfranz - can you take a look at this? Also I am not sure how to handle ensuring that the app reads the newly downloaded file.

maxkfranz:
@jvwong Let's review in-person tomorrow, if you're available.  We can finalise the code then

Issue #1234, opened 2018-12-13
Interactions gene tooltip: Add summary information 


jvwong:
## Description of new feature

The interactions node tooltip simply displays  dummy content - a hardcoded link to HGNC symbol page. Add entity summary information. 

## Motivation for new feature

Add useful information about a node: Names, description and links.

jvwong:
You can take it? Can expose our geneSearch function?

PR #1252, opened 2018-12-17
lazy load interactions node info

d2fong:
refs #1234
screen shot 2018-12-17 at 3 20 58 pm


d2fong:
ok

jvwong:
while you're at it, let's remove the endpoint `/api/summary/entity/search` and function `entitySearch` as they will be unused.

Issue #1196, opened 2018-11-30
When it shows a Factoid document by id, download feature is broken


IgorRodchenkov:
E.g., http://appsbeta.pathwaycommons.org/factoids/86f7fc9b-2a91-4cbb-9320-1a3e98a3abc5
- cannot download as BioPAX, SIF, SBGN, etc. I think, this is broken due to re-using the menu components and buttons that were originally made for network views/data generated from PC search (i.e., "normal" app-ui use case); then, we connected this and factoid apps and should apparently call different services (factoid api/biopax/{id}, api/sbgn/{id}) to get the downloads work.

IgorRodchenkov:
@d2fong honestly, i am not sure how to do this. 
Basically, the links what show up once you click in the left-top menu button, in the case of BioPAX and SBGN-ML downloads, should use the Factoid api/document/biopax/{docId} or ../sbgn/{docId} api instead of what/where they currently do/go. However, this change should not break the download actions for the normal use case; i.e., when we go from app-ui search to viewing a pathway or interactions (with no factoid involved)... Could you look into this, please?

jvwong:
Two options:
1. Remove or filter out those non-functional
2. Disable

PR #1277, opened 2019-01-23
Factoid download options

jvwong:
Added a general option to explicitly disable certain 'pc-download-types' in the toolbar file-download-menu button. 

Factoid looks like this:

image

Refs #1196

### DIFF
--- a/src/client/features/pathways/index.js
+++ b/src/client/features/pathways/index.js
@@ -71,6 +71,7 @@ class Pathways extends React.Component {
 
   render() {
     let { loading, pathway, cySrv, networkEmpty, error } = this.state;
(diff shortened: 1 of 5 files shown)


maxkfranz:
Would it be better to remove those options?  It might be confusing to show disabled options without showing a reason why they're disabled.

jvwong:
In general, it makes more sense to remove and require explicit instructions to add a file type. For Factoid, probably 50/50 so will just do above...

jvwong:
Now the Factoid case displays the default ('png') and the Pathways app shows all the download types by declaring each explicitly. The 'Interactions' and 'Enrichment' app use their own components, so I left those alone.

Factoid:

image

Issue #1289, opened 2019-01-28
Update Docker 


jvwong:
- Remove database
- Point to master

PR #1293, opened 2019-01-28
Docker updates

jvwong:
Minor Docker updates. 

Refs #1289

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,7 +21,7 @@ RUN npm install
 
 # Build project
 RUN npm run clean
-RUN npm run build-prod
+RUN npm run build
 
 # Expose port
 EXPOSE 3000


Dockerfile
```diff
@@ -19,16 +19,16 @@ COPY . /home/appuser/app
 # Install app dependencies
 RUN npm install
 
-# Build project 
+# Build project
 RUN npm run clean
-RUN npm run build
+RUN npm run build-prod
```
maxkfranz: I'd rather leave it how it was before.  This makes it impossible to ever build the Docker with debuggable code.

From the readme:

> Build the container. Here, app-ui is used as the container name.
>
>cd app-ui
>docker build --build-arg "NODE_ENV=production" -t app-ui .
>
>Run the container:
>
>docker run -it -p 12345:3000 -u "node" -e "NODE_ENV=production" --name "app-ui" app-ui

Basically, it's up to the consumer of the dockerfile to decide what env vars he wants to use to configure the image.

jvwong: ok

Issue #1221, opened 2018-12-11
Swap out xref2uri implementation


jvwong:
## Description of new feature

The server offers an external service for pc that maps a (MIRIAM) [name, record] tuple to a resolvable URI for purposes of web link outs (e.g. 'for more info' about a gene). The `xref2uri` function uses a web service that accepts [name, record] pairs one-by-one.

Goal is to use http://biopax.baderlab.org/docs/index.html that can accept multiple queries per call.

## Motivation for new feature

Much lower overhead (single HTTP) for calls to map to URI.

jvwong:
NB: `pc2/miriam/uri/` going away in v11.

PR #1304, opened 2019-02-01
Using the BioPax Suggester implementation

jvwong:
Swaps out the web service endpoint underlying `xref2Uri` that maps an xref (db, id) to URI. In particular, making *minimal* changes to replace  `http://www.pathwaycommons.org//pc2/miriam/uri/{db}/{id}` for `http://biopax.baderlab.org/xref/`.

Refs #1221.


jvwong:
@IgorRodchenkov made this service for the apps - looks like the one we're using now will go away in v11 / beta.

IgorRodchenkov:
@maxkfraz 
http://biopax.baderlab.org/docs/index.html
will stay bacause it is part of the BioPAX Validator app

Issue #1312, opened 2019-02-04
(One last) Fallbacks for fetch errors


jvwong:
Prevent app 'hanging' by providing reasonable fallbacks for fetch errors in: 
 
- pathways (paint)
  - menus 
    - Downloads: `downloadFileFromPC` 
- interactions
  - edge-tooltip
    - `getPublications`

PR #1313, opened 2019-02-04
Show error message on file download fetch error.

jvwong:
Instead of hanging spinner, download menu displays error message.
Refs #1312


jvwong:
Yeah i've never seen it and I have to rig these fail cases. I just don't like infinite spinners.

PR #1315, opened 2019-02-04
Interactions tooltip error fallback

jvwong:
Interactions tooltip: Simply bypass errors encountered when attempting to fetch publications and show minimal data. 
Refs #1312


jvwong:
on error (skip publications section):

image

Issue #1316, opened 2019-02-04
Warning in Interactions edge selection: setState in render


jvwong:
Edge tooltip calls a state change within the render() method. Remove.

PR #1317, opened 2019-02-05
Interactions tooltip: Remove state calls in render

jvwong:
Cleanup: Set initial state during construction/mounting rather than during render().
Refs #1316.

### DIFF
--- a/src/client/features/interactions/interactions-edge-tooltip.js
+++ b/src/client/features/interactions/interactions-edge-tooltip.js
@@ -9,11 +9,13 @@ class InteractionsEdgeTooltip extends React.Component {
   constructor(props){
     super(props);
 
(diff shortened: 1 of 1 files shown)


src/client/features/interactions/interactions-edge-tooltip.js
```diff
@@ -9,13 +9,22 @@ class InteractionsEdgeTooltip extends React.Component {
   constructor(props){
     super(props);
 
+    const edges = props.edge.parallelEdges();
```
maxkfranz: Preferably, graph topology calls shouldn't be made at all in `render()` -- because `render()` can be called many times.  Since react can destroy and re-create components, it would be even better to raise the parallel edges call higher up in the component hierarchy.  But that could be considered for some future version; this is good enough for now.

Issue #1251, opened 2018-12-17
Show interactions card only if there is data


d2fong:
1. enter the following query in the search app:
ADAM10	ADAM17	AGXT	ANGPT4	ANXA4	APH1A
2. click the interactions app linkout
3. see that there are no interactions to display
4. this should not happen

d2fong:
Solution:
- prefetch the interactions/enrichment networks and check if they are empty

Two approaches:
1. the server can do this to reduce complexity of the front end, the front end then checks the server response to see which app linkouts to display
2. the front end can fetch both networks and decide itself which one to display, absolving the server of that responsibility

Current thinking: let the frontend handle it

jvwong:
Relates to case when there are only a few genes and enrichment analysis might not be sensible.

- Alternatives
  - always show (current)
  - hide card
  - disable card

jvwong:
e.g. TMCC2 FUT10 GPR65 STPG2 PIRC91 PRDX5 (TP53)

PR #1323, opened 2019-02-06
Show interactions card only when data exists

jvwong:
The app cards already have a `show` flag for display. This PR provides a means to 'peek' at the interactions data in order to set the interaction card `show`.

Refs #1251.

Sample data: TMCC2 FUT10 GPR65 STPG2  (TP53)

### DIFF
--- a/src/client/features/search/gene-results-view.js
+++ b/src/client/features/search/gene-results-view.js
@@ -4,7 +4,9 @@ const { Link } = require('react-router-dom');
 const queryString = require('query-string');
 const _ = require('lodash');
 
+const { ServerAPI } = require('../../services');
 const MIN_GENE_COUNT_ENRICHMENT = 5;
+const MINIMUM_CONTENT_LENGTH = 35;
 const { NS_HGNC_SYMBOL, NS_GENECARDS, NS_NCBI_GENE, NS_UNIPROT } = require('../../../config');
 
 const SUPPORTED_COLLECTIONS = new Map([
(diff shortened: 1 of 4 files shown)


maxkfranz:
I'm on the fence about this approach:  While it prevents the card showing up for an empty graph, it adds nontrivial expense to the search page.  

On the UX side, a user may reasonably expect the card to always be there.  If there are no edges for a set of nodes, a user could expect to see the nodes without any edges after clicking the card -- or at least a warning message.  (The ideal would be a graph with only nodes and an accompanying warning message.)  The absence of the edges can be useful information, and it may be better to be explicit about that rather than just hiding the card.  

To a user, it looks like it's broken when the card is missing *sometimes* and the reasoning isn't obvious. The enrichment case seems much more obvious than the interactions case.

Even for enrichment, I'd rather grey out the card when disabled rather than hiding it.

maxkfranz:
Having a greyed out enrichment card would also help with discoverability:  Most users enter a single gene query.  Those users never see the enrichment card, so they may never know about it.  They may not even know that they can enter more than one gene.

So by having a greyed out enrichment card with an appropriate message -- something like "add more genes to your search to get an enrichment result" -- more users could organically discover the enrichment app.

jvwong:
> While it prevents the card showing up for an empty graph, it adds nontrivial expense to the search page.

As far as I can tell, pathway list rendering doesn't depend on the genes/cards component. The interactions peek is pretty quick, but there could be some tiny flicker on re-render. Doing the peek on the server and putting it in search results would indeed tie the whole thing together. 

> On the UX side, a user may reasonably expect the card to always be there...Having a greyed out enrichment card would also help with discoverability...

- What about:
  - app cards have flag: `enabled`
  - flag sets the card class: `enabled: false` means greyed out, doesn't link out
  - peek as before to update `enabled`
    - Ignore enrichment for now, as it's a *very* heavy, slow call.

maxkfranz:
I would modify that approach by not doing any additional queries:  Interactions is always enabled and enrichment gets toggled based on the simple rule we already have.  We just need to edit the interactions result page a bit.

jvwong:
Looks like this: 

image

Issue #1328, opened 2019-02-14
Update to accommodate changes to g:GOSt GMT download file format and names


jvwong:
Background: The enrichment modules download and use a GMT file from g:GOSt. As well, this file is automatically updated via a cron job. 

g:Profiler has recently updated their site. As part of this, the name and format of the GMT files have been altered. In particular, GMT files are now separated by data source (i.e. GO:BP, Reactome) whereas previously, they were combined.

Update to accommodate these changes.

jvwong:
![image](https://user-images.githubusercontent.com/4706307/52806216-6ae19480-3056-11e9-8702-2543062fd9e7.png)

jvwong:
update tasks:

- [ ] ability to pick not just one but multiple GMT files from zip archive
- [ ] concatenate results

jvwong:
NB: `/` is reserved character so it's mapped to `:` in `hsapiens.GO/*.name.gmt`

PR #1329, opened 2019-02-19
Accommodate changes to g:GOSt GMT download file format

jvwong:
g:Profiler g:GOSt has updated their GMT data file formats and names. Technically, the production app contains an out-of-date GMT file so visualization might not match analysis output.

- Pick multiple GMT files from zip 
- Concatenate to final GMT file
- Misc test updates 

Refs #1328


jvwong:
@maxkfranz can you take a look. I think this should be in next release.

jvwong:
`GPROFILER_URL` is the same (`https://biit.cs.ut.ee/gprofiler/`)

Issue #1322, opened 2019-02-06
Enrichment: Error on click 'Reset Arrangement' 


jvwong:
## Environment info

- Browser version (client) : Firefox (), Chrome (Version 72.0.3626.81)
- Node.js version (server) : 8.11.2, 10.15.1

## Current (buggy) behaviour

On click 'Reset Arrangement' menu button, there is a console.error: 

```
Can not create second element with ID `component-0` cytoscape.cjs.js:296
```

jvwong:
Related: cyjson/download outputs parent node as `component-i`.

PR #1330, opened 2019-02-20
Add parent node once only.

jvwong:
Refs #1322.


jvwong:
Oh yeah, that's no good. OK talk later.

Issue #1359, opened 2019-04-23
Enrichment validation: Optional target database param isn't optional. 


jvwong:
## Current (buggy) behaviour

The enrichment validation service `targetDb` parameter is documented as optional and defaults toHGNC ID. It is in fact required. 

## Desired behaviour

Should be optional and default to HGNC ID.

## Minimum steps to reproduce

See docs http://appsbeta.pathwaycommons.org/api/enrichment/docs#/Validation_Service/geneQuery

PR #1360, opened 2019-04-30
Make validation target database optional  

jvwong:
Renamed the target database parameter selection to `namespace` and made it optional; defaults this value to `hgnc`. 
Refs #1359


maxkfranz:
We should change the travis config to test node 8 and 10.  'stable' has bumped it up to 12, which probably won't work...

PR #1366, opened 2019-06-16
Batch node search, count (pathways and SIFs) script

IgorRodchenkov:
- this is e.g., to run all those query strings from a Reactome's monthly "failed user's search" report through our PC apps flow and compare, analyse, celebrate or cry the differences...

### DIFF
--- a/src/scripts/batch/hit_counts.js
+++ b/src/scripts/batch/hit_counts.js
@@ -12,12 +12,12 @@
  * @author Igor Rodchenkov
  */
 
+global.fetch = require('node-fetch');
 const {getInteractionGraphFromPC} = require('../../server/routes/interactions/generate-interactions-json');
 const {search} = require('../../server/routes/search/search');
 const csv = require('csv');
 const {pipeline} = require('stream');
 const logger = require('../../server/logger');
-
 
 const transform = csv.transform((data, callback) => {
   let out = [];
--- a/src/scripts/batch/hit_counts.js
+++ b/src/scripts/batch/hit_counts.js
@@ -1,0 +1,100 @@
+/**
+ * Using user query strings from the input stream,
+ * this script gets the number pathways and interactions (SIF edges).
+ * Results will be the same as or close (see comments below) to
+ * what one could get at apps.pathwaycommons.org if settings and queries are same.
+ *
+ * This script simply calls the app-ui server-side modules
+ * (does not start the nore server nor web app),
+ * which, in order, search for pathways and inferred simple
+ * interactions in the Pathway Commons backend (PC2 and Sifgraph services).
+ *
+ * @author Igor Rodchenkov
+ */
+
+const {getInteractionGraphFromPC} = require('../../server/routes/interactions/generate-interactions-json');
+const {search} = require('../../server/routes/search/search');
+const csv = require('csv');
+const {pipeline} = require('stream');
+const logger = require('../../server/logger');
+
+
+const transform = csv.transform((data, callback) => {
+  let out = [];
+  let q = data['Term'];
+  out.push(q);
+
+  // apply the same hack as done on the app-ui's client-side, in ServerAPI.search (before  going /api/search)
+  if (/^((uniprot|hgnc):\w+|ncbi:[0-9]+)$/i.test(q)) {
+    q=q.replace(/^(uniprot|ncbi|hgnc):/i,"");
+  }
+
+  // A quick-fix and note -
+  // PC's all-in-one full-text index (unlike name: and xrefid: fields) was build
+  // using Lucene StandardAnalyzer, which treats '-' (dash), etc., as wildcard when tokenizing original text,
+  // which, e.g., results in q="R-MMU-975956" matches too many and irrelevant pathways (it's like R OR MMU OR 975956).
+  // One should search like q=name:R-MMU-975956 or q=xrefid:R-MMU-975956 or q="R-MMU-975956" (i.e., quotes are part of the query),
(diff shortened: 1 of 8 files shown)


src/server/external-services/pathway-commons.js
```diff
@@ -140,7 +140,7 @@ const sifGraph = opts => {
   return fetch( url, { method: 'GET', headers: { 'Accept': 'text/plain' } } )
   .then( res => res.text() )
   .catch( e => {
-    logger.error( 'sifGraph ' + opts + ' failed - ' + e );
```
IgorRodchenkov: opts did not work here

src/scripts/batch/hit_counts.js
```diff
@@ -0,0 +1,100 @@
+/**
+ * Using user query strings from the input stream,
```
IgorRodchenkov: here we go

src/scripts/batch/hit_counts.js
```diff
@@ -0,0 +1,100 @@
...
+  let out = [];
+  let q = data['Term'];
+  out.push(q);
+
+  // apply the same hack as done on the app-ui's client-side, in ServerAPI.search (before  going /api/search)
+  if (/^((uniprot|hgnc):\w+|ncbi:[0-9]+)$/i.test(q)) {
+    q=q.replace(/^(uniprot|ncbi|hgnc):/i,"");
+  }
+
+  // A quick-fix and note -
```
IgorRodchenkov: FYI -

src/scripts/batch/hit_counts.js
```diff
@@ -0,0 +1,100 @@
...
+const {pipeline} = require('stream');
+const logger = require('../../server/logger');
+
+
+const transform = csv.transform((data, callback) => {
+  let out = [];
+  let q = data['Term'];
+  out.push(q);
+
+  // apply the same hack as done on the app-ui's client-side, in ServerAPI.search (before  going /api/search)
```
IgorRodchenkov: FYI - see comments below

src/util/fetch.js
```diff
@@ -1,5 +1,6 @@
 const { promiseTimeout, TimeoutError } = require('./promise');
 const _ = require('lodash');
+const fetch = require('node-fetch');
```
jvwong: This duplicates the 'node-fetch' in `server/index.js`

jvwong: This is also the reason the tests are failing, because it stomps on my mock server response. Remove.

src/scripts/batch/hit_counts.js
```diff
@@ -0,0 +1,100 @@
...
+  if (/^((uniprot|hgnc):\w+|ncbi:[0-9]+)$/i.test(q)) {
+    q=q.replace(/^(uniprot|ncbi|hgnc):/i,"");
+  }
+
```
jvwong: Yeah I should probably ditch this in the client as we don't use it anymore.

src/scripts/batch/hit_counts.js
```diff
@@ -0,0 +1,100 @@
...
+      out.push(searchHits.length);
+      return genes.map(geneInfo => geneInfo.geneSymbol);
+    })
+    .then((symbols) => {
+      if (symbols && symbols.length > 0) {
+        return getInteractionGraphFromPC(symbols);
+      }
+    })
+    .then((graph) => {
+      (graph && graph.edges) ? out.push(graph.edges.length) : out.push(0);
```
jvwong: In this case of one query gene, number is minimum number of edges since we cap at 25.

IgorRodchenkov:
The script doesn’t work without node-ferch reqire in the fetch.js (where it should be I bet, because ‘fetch()’ is there called but wasn’t imported. If requiring node-fetch in fetch cause issues and tests fail, then we likely have app/test design issues...

IR

jvwong:
This breaks the app/front-end too.

src/util/fetch.js
IgorRodchenkov: @jvwong this change should not break anything.. there is something wrong in the app, perhaps a bigger hidden issue... I guess, we should not probably use fetch.js file name and import it as "const fetch=require('../../fetch'") where it is used.. or use another name for the const...

src/scripts/batch/hit_counts.js
```diff
@@ -0,0 +1,100 @@
...
+ * what one could get at apps.pathwaycommons.org if settings and queries are same.
+ *
+ * This script simply calls the app-ui server-side modules
+ * (does not start the nore server nor web app),
+ * which, in order, search for pathways and inferred simple
+ * interactions in the Pathway Commons backend (PC2 and Sifgraph services).
+ *
+ * @author Igor Rodchenkov
+ */
+
```
jvwong: Suggestion is to insert `global.fetch = require('node-fetch');` as in server/index.js  instead of below.

src/scripts/batch/hit_counts.js
IgorRodchenkov: Why.. this was in the client to support linking from UniProt (eventually). Let it be. I copied this here because the script is suppose to work alike the UI, and the Reactome users did type something like "uniprot:q12345", which would not match anything in our app without the above code, i think.

src/scripts/batch/hit_counts.js
IgorRodchenkov: This code is not about one query gene; this just counts graph.edges.length of whatever query returned (query could be any text - from the first column each row in input file - might contain a word, phrase with multiple genes, etc. that users typed at Reactome's site)

src/scripts/batch/hit_counts.js
IgorRodchenkov: done

Issue #1374, opened 2019-09-27
v12/Beta: Bugs


jvwong:
Interactions app is failing - 'An error occurred'

Reproduce:
- [search `TP53 MDM2 CDKN1A ZNF385A`](http://appsbeta.pathwaycommons.org/interactions?source=TP53%2CMDM2%2CCDKN1A%2CZNF385A)
- response `An error occurred`

Problem:

Our old hardcoded 'datasources.js' file is rearing its ugly head: Missing a new datasource (pathbank). Origin of the problem: data provider names in the sif file are english text, not uri, provenance etc, hence the hardcoding and not using the 'getdatasources()' function.

jvwong:
Pathway viewer downloads are all empty.

e.g. http://appsbeta.pathwaycommons.org/pathways?uri=http%3A%2F%2Fidentifiers.org%2Fsmpdb%2FSMP0031692

PR #1375, opened 2019-09-27
Remove hard-coded data provider file  

jvwong:
There was some technical debt regarding the fetch and presentation of database provider data (name, url, etc). In particular, we were using a hand-coded json file of providers in `src/models/datasources.js`. This broke our app when a new provider (Pathbank) was added. This PR fetches this information dynamically from a pc service. 

- removed hand-coded `src/models/datasources.js`  of data providers
- Pathway model class obtains provider homepage url from server-side data
- interactions app finds edge provider names from pc service
- pathways api fetches data provider name and url from pc service 

refs #1374


maxkfranz:
Looks good.  Async/await makes this more readable.

Issue #1378, opened 2019-10-03
Refactoring: enrichment map pathway file downloads


jvwong:
There were several outstanding issues with the enrichment module and service:

- Uses a checked-in file for GMT [bug]
  - When the app starts, it is using GMT data that could potentially be out of date. The cron job runs Mondays, so you could be using an out of date file for a week. 
  - Should refresh the GMT data on boot
- Module to download and write archive file can be reused
  - This module could be updated to be reusable anytime one wishes to pick off an external compressed archive and have it locally (e.g. generic entities, GMT data, ...)
  - Make reusable module
- The tests are fragile, wrong
  - Too specific to a particular data/pathway values; some were plain wrong
  - Write  tests that are not data specific, test existence of keys

jvwong:
Additional: see https://github.com/PathwayCommons/app-ui/pull/1379#issuecomment-540107011

PR #1379, opened 2019-10-04
Refactoring a module for downloading archive files

jvwong:
I factored out a more reusable module in `source-files.js` that simply downloads an archive, picks off the files you care about, and writes them to the host.

The enrichment visualization module was updated to use this function, and so downloads a fresh set of GMT file from source and concatenates them on app start into `pathways.gmt`. The existing (stale) file `pathways.gmt` was removed from version control. Rest is tests.

Refs #1378


jvwong:
> I'm not enthusiastic about storing generated files under src. It makes it seem like the file should exist as soon as you clone the repo.

Are you saying that it's odd  that `pathway-table.js` reads in `GMT_SOURCE_FILENAME ` that doesn't exist in the repo?

maxkfranz:
No, it's weird that a generated file exists under `src`.  It would be better for it to exist under some top-level `downloads` directory.  Imagine you're someone who clones the repo and just started working on the project with no prior knowledge.  That there are files missing under `src` seems like someone forgot to check something in.

Unless we're going to clean up all cases like this now, let's just merge this in and make a separate issue for future.

jvwong:
This one's already pretty big including tests etc. I'll merge then think about improving this one.

Issue #1269, opened 2019-01-17
Improve node filtering criterion for interactions app


maxkfranz:
TODO...

## Environment info

- Browser version (client) :
- Node.js version (server) :

cannin:
Some nodes (or alter) that might be nice to remove: 

## Hide the name "noname": 
* http://appsbeta.pathwaycommons.org/pathways?uri=http%3A%2F%2Fidentifiers.org%2Fwikipathways%2FWP254

## Interactions that terminate in a process node: 
* http://appsbeta.pathwaycommons.org/pathways?uri=http%3A%2F%2Fidentifiers.org%2Fwikipathways%2FWP254

<img width="139" alt="screen shot 2019-01-23 at 11 57 07 am" src="https://user-images.githubusercontent.com/790284/51623389-96b1a480-1f06-11e9-8bd8-2e77d57fc15f.png">

jvwong:
For neighbourhood: Prioritize sourced from Biofactoid. These would also, generally be the most 'recent'.

PR #1402, opened 2022-06-29
Biofactoid-derived interactions

jvwong:
The goal here was to bias the interactions app to show interactions from Biofactoid, which, in all likelihood will be the most recently identified anyways.

This PR is rather heavy-handed (i.e. just pushes biofactoids to the top of the pile), and maybe there's other ways? 

Refs #1269


maxkfranz (approved):
Great

Issue #1401, opened 2022-06-28
Referrals to Biofactoid 


jvwong:
## Description of new feature

As of Pathway Commons version 13, Biofactoid pathways should be available and interactions between human genes/products should be available. This provides an ideal opportunity to raise awareness of Biofactoid and refer traffic to the website for those interested in exploring in more detail and possibly contributing their own information, that will end up in subsequent PC updates.

The feature could be as a simple link on the search landing page:

![Search-Results](https://user-images.githubusercontent.com/4706307/176462341-cc4aa9be-92be-4e7d-9f2a-97e298dcac1e.png)

## Motivation for new feature

The PC apps see a reasonable amount of traffic on a daily basis, presumably from researchers that are looking for pathway and interaction information. Table 1 gives an overview of the users and sessions on a daily basis so far in 2022: 

<img width="1417" alt="Screen Shot 2022-06-28 at 4 22 14 PM" src="https://user-images.githubusercontent.com/4706307/176278278-844048b5-7d85-4dda-8d94-ed8407beb7ae.png">

A substantial amount of referral traffic is originating from either UniProt (48% sessions) and GeneCards (13%). For those that land on the search, a good proportion (36%) follow through by viewing the interactions app, and of those, nearly 45% circle back to search again, presumably to refine or search again:

<img width="1340" alt="Screen Shot 2022-06-28 at 4 22 18 PM" src="https://user-images.githubusercontent.com/4706307/176278285-063f24a8-a934-4938-b26d-a9ba21f133e5.png">

jvwong:
> It would be nice to link directly to the Biofactoid viewer in search results or at least the pathway viewer equivalent -- or maybe even both. Maybe the factoids aren't a good fit under the 'Pathways' list, but it would be nice to see individual factoids as search results even if they're in a different section at the top, like 'Biofactoids'. The ideal would be for Biofactoid results to be card results with pictures of the figures, but text results like the 'Pathways' list would be good too.

An idea follows from a special case: If the top hit under 'Pathways` comes back from source biofactoid, then you can privilege this hit with (a) images (b) related text (c) related links. This is borrowed from Google's contextualized results:



<img width="800" alt="Screen Shot 2022-07-05 at 12 27 30 PM" src="https://user-images.githubusercontent.com/4706307/177374549-b833ef61-b9ae-43fb-83de-7b96c44e9ee6.png">

jvwong:
**Case I: Query is the top hit in pathway search**
[...]

PR #1410, opened 2022-09-19
Feature panel

jvwong:
A feature panel, that displays a summary of a Biofactoid paper/pathway, when it comes up as the first pathway hit. 

Hover to show document 'context' (when available) and interactions text.

- Questions
  - Is this the right way to highlight/sell Biofactoid?
  - Kinds of info to add/remove?
- Notes
  - Local instance set environment `"PC_URL": "https://beta.pathwaycommons.org/"`
  - Refs #1401
  - Performance: Trivial hit - most time spent on backend is getting gene info from gProfiler 

#### Mixed species

Here, cofilin is human, but rest are mouse


jvwong:
Sorry wrong target branch

PR #1411, opened 2022-09-19
Feature panel

jvwong:
A feature panel, that displays a summary of a Biofactoid paper/pathway, when it comes up as the first pathway hit. 

Hover to show document 'context' (when available) and interactions text.

- Questions
  - Is this the right way to highlight/sell Biofactoid?
  - Kinds of info to add/remove?
- Notes
  - Local instance set environment `"PC_URL": "https://beta.pathwaycommons.org/"`
  - Refs #1401
  - Performance: Trivial hit - most time spent on backend is getting gene info from gProfiler 

#### Mixed species

Here, cofilin is human, but rest are mouse


jvwong:
Re:  discussion of modifications

- Diminish the separation of paper and Biofactoid data
  - reconsider separator
  - Drop subtitles (Pathways, Genes)
- Make Biofactoid more prominent and obvious as a section
  - Bold a subtitle instead of credit footer (like "Genes Detected")
  - Card above/beside paper 
- Unify information within card
  -  Genes/Chemicals can be hyperlinked with source logo
  -   species in hover
  - scrollable content to accomodate extra text

jvwong:
I updated the Feature View to create more unity between the article and the Biofactoid. One good example and then a problem below:

#### A good example ('TNL1')

Screen Shot 2022-09-22 at 5 57 09 PM

#### A problem with the AppCard content

I tried to make the AppCard content a sibling of the image, as it complained of nested anchors when there were links in the content. However, when there's a lot of text, the content translates so far as to cover the image and precludes one from clicking the image to go to Biofactoid:
 
Screen Shot 2022-09-22 at 5 57 33 PM

Any ideas?

jvwong:
Tried making the image text-panel scrollable:

1) Typical case (minimal content)

scoll_typical

2) Edge case (lots of content)

I leave the top of the image visible, text-panel logo/species block is hyperlinked.

Screen Recording 2022-10-06 at 6 40 43 PM

maxkfranz:
The first thing to try is to ensure that the max-height of the overlay is 100% of the card height, so it can’t become taller than the card.  N.b. the parent card div must be position: relative for percent sizing of the child (overlay) to be relative to the parent (card) size. You may also need overflow-y:auto on the overlay to make it scrollable.If that doesn’t work, let me know and I’ll play with it. On Sep 22, 2022, at 18:07, Jeffrey ***@***.***> wrote:﻿
I updated the Feature View to create more unity between the article and the Biofactoid. One good example and then a problem below:
A good example ('TNL1')

A problem with the AppCard content
I tried to make the AppCard content a sibling of the image, as it complained of nested anchors when there were links in the content. However, when there's a lot of text, the content translates so far as to cover the image and precludes one from clicking the image to go to Biofactoid:

Any ideas?

jvwong:
What do you think of the last commits/screenshot - not pretty but I think it works OK.

Issue #1409, opened 2022-09-19
Escape special chars (Lucene) in queries


jvwong:
Lucene interprets queries with certain characters (e.g. slash) as special (e.g. regex). This leads to some HTTP code 500 errors.

Reproduce:
- [search: 'AKT/AMPK'](https://www.pathwaycommons.org/pc2/search?type=Pathway&q=AKT\/AMPK)
  - thinks its a regex 
- [search: 'Hsp70 Interacts with Mitogen-Activated Protein Kinase ('](https://www.pathwaycommons.org/pc2/search?type=Pathway&q=Hsp70%20Interacts%20with%20Mitogen-Activated%20Protein%20Kinase%20%28MAPK%29-Activated%20Protein%20Kinase%202)
  - not sure, maybe some OR logical... 
 
Can simple escape these (https://lucene.apache.org/core/2_9_4/queryparsersyntax.html#Escaping%20Special%20Characters). Not sure if at some point, we allowed Lucene query language through the UI but probably shouldn't 

References

- https://lucene.apache.org/core/4_0_0/queryparser/org/apache/lucene/queryparser/classic/package-summary.html#Regexp_Searches

PR #1412, opened 2022-09-19
Escape special characters

jvwong:
This [lucene-escape-query](https://www.npmjs.com/package/lucene-escape-query) seems to take care of most cases that break the pc server. However, doesn't appear to escape slashes so things like `AKT/AMPK` return nothing unlike `AKT\/AMPK`. 

Refs #1409

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6262,6 +6262,11 @@
         "yallist": "^2.1.2"
       }
     },
+    "lucene-escape-query": {
+      "version": "1.0.1",
+      "resolved": "https://registry.npmjs.org/lucene-escape-query/-/lucene-escape-query-1.0.1.tgz",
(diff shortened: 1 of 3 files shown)


maxkfranz:
If there isn't support for slashes in the library, would it be reasonable to handle some common cases manually?  How frequent would queries with slashes, like 'AKT/AMPK', be?  Are there other cases worth considering now (e.g. maybe dashes)?

Some possibilities:

(1) Just strip out some problematic characters (either delete or replace with something else that is supported, like a space).  That's something like `.replace('/[/|\]/g')` if `/`, `|`, and `\` are problematic.  The search might function well enough by just removing some of these characters.

(2) Transform specific characters for escaping manually.

Issue #1415, opened 2022-10-21
Component tweaks


jvwong:
A. Feature Component
Article title should hyperlink --> Biofactoid document

B. App Card
When no hyperlinks appear within sliding panel (on hover), the panel effectively blocks one from navigating to the target site (e.g. the interactions, enrichment)

PR #1417, opened 2022-11-01
App card should linkify the sliding panel for interactions & enrichment.

maxkfranz:
> B. App Card
> When no hyperlinks appear within sliding panel (on hover), the panel effectively blocks one from navigating to the target site (e.g. the interactions, enrichment)

The feature / Biofactoid card should be unaffected.

Ref: Component tweaks #1415 (B)


jvwong (approved):
Great thanks!

Issue #1425, opened 2022-11-28
Feature: Top pathway hit


jvwong:
**Goal**
Feature a the top pathway search hit and include a visual preview (SBGN) when a query references at least one of the molecular participants. 

**Primary use case**
Referral from public entity databases. In these cases, the query will be in the form of a (a) UniProt accession (referrer: UniProt) and/or b) HGNC symbol (referrer: GeneCards). The response should contain information necessary to populate a Feature Component for the pathway: name, source, #participants, and a URL pointing to an SBGN preview]

**Restrictions & Limitations**
Main concern is slowing down the "time to interactive" of the search, which degrades the user experience. Several potential  bottlenecks:
  - Generating an SBGN snapshot on-the-fly
  - Identifying entity IDs from the query
  - Retrieving a pathway's participants (UnificationXref db, id).

**Edge cases**
Integrating with case where Biofactoid is contained in a Feature view (i.e. triggered when top search hit source is Biofactoid)

**Notes**
To reduce bottlenecks, might consider

a. Rendering and caching SBGN previews for all pathways instread of on-the-fly, which might also be useful in Google indexing
  - https://journals.plos.org/ploscompbiol/article?id=10.1371/journal.pcbi.1010635
b. Have cPath2 index and return UnificationXrefs to bypass need to retrieve these on-the-fly information separately

jvwong:
### Update

**Goal**
Feature a the top pathway search hit and include a visual preview (SBGN) so that users are aware of what the pathway section links will hold.

**Primary use case**
Any user that visits the search page.

**Mockup**

![summary](https://user-images.githubusercontent.com/4706307/204637850-614b4981-b56e-41e0-97e3-a39a3f342fd6.png)

**Details**

Benefits:
- Generic to all users of the search, whether referred (UniProt, GeneCards) or arriving via Google
- Does not change the experience for majority of users who are mainly interested in interactions app 
- Preview is placed in the Pathway section, indicating that other links will provide the same info (SBGN)
- No complex mapping of query to pathway previewed
- User existing components (e.g. SBGN pathway viewer)

Risks:
- The top hit may or may not be relevant (i.e. we're not checking that a database ID query is in the pathways)
- Generating the SBGN visualization may be slow
- SBGN network can be enormous, complex

cannin:
Python code to generate static images of PC pathways using the Syblars (https://pubmed.ncbi.nlm.nih.gov/36374853/ from Ugur Dogrusoz)

[...]

PR #1443, opened 2024-01-10
A command-line tool to generate SBGN images for PC pathways 

jvwong:
This provides command line tools to generate a collection of image files of SBGN pathway representations, say from a list of pathways declared in a GMT file available from Pathway Commons. This would useful in providing PC pathway previews in the Search app, a PC pathway static site and/or improving the chance that pathway images are indexed by Google and picked up in user searches.

## Details

- Command line programs
  - `source`: Download and extract a file
  - `snapshot`: Generate PNGs for pathways listed in a PC GMT-formatted file
- Dependencies
  - [Syblars](http://syblars.cs.bilkent.edu.tr/)
    - Setup provided herein by `Dockerfile.syblars` and `docker-compose.yml`
- Configuration
  - `SBGN_IMG_SERVICE_BASE_URL`: Location of Syblars instance
  - `SBGN_IMG_PATH`: Where to save image files (default: `public/img/pathways`)
  
**Performance**

I did a test run using pathways in [PathwayCommons12.All.hgnc.gmt.gz](https://www.pathwaycommons.org/archives/PC2/v12/PathwayCommons12.All.hgnc.gmt.gz) on my system (iMac 3.6 GHz 8-Core Intel Core i9) with a local Docker instance of Syblars outputting PNG files.

- Results
  - N=3971
  - Storage  requirements:  Directory 744 MB
  - Elapsed processing time: 73 Minutes

### Notes

- Used the PC pathway URI as the filename, but had to mangle the URI to something legal for a file name (i.e. replace symbols with `_`). There may be a better way to handle this so there's no mapping needed, possibility of clashes.
- `snapshot` is configured to dump images to `public/img/pathways` so expressJS can serve it at `img/pathways/:id`, where `id` is the mangled pathway URI

### Syblars

Getting Syblars to work locally required some fiddling, hence the choice to include the `Docker.syblars`. I also suspect a leak, as it would gradually hold onto RAM while processing images, crash with an out of memory error (> 4096 MB configured), then reboot clean:
 
docker_memory

## Future considerations

- Incremental updates
- Rolling our own sbgn-2-image generator (see https://github.com/PathwayCommons/app-ui/issues/1425#issuecomment-1854392914)
- A pathway data file, with image file name and associated metadata (genes, source, etc) which is unused currently. 

## Related
  
  - #1425 
  - https://github.com/nrnb/GoogleSummerOfCode/issues/218 
  - https://github.com/PathwayCommons/app-ui/issues/1425#issuecomment-1854392914


maxkfranz:
Is Chromium/Syblars itself configured to use a maximum amount of memory?  Even without a leak, browsers can be aggressive with allocating memory, because UI end-users don't like it when their tabs slow down or crash.

jvwong:
> Is Chromium/Syblars itself configured to use a maximum amount of memory? Even without a leak, browsers can be aggressive with allocating memory, because UI end-users don't like it when their tabs slow down or crash.

Not sure. I didn't customized either. 
I ran this again with a Syblars container using the (mostly) original Dockerfile and the result was the same: It holds onto memory until it hits the default NodeJS limit (~2 GB) then the container exits.

IgorRodchenkov:
We could use the md5 hash of a pathway URI (could still add the sanitized original URI as preix) for the file names @jvwong

jvwong:
> We could use the md5 hash of a pathway URI (could still add the sanitized original URI as preix) for the file names @jvwong

Right yes - during development I wanted to look up specific pathways by their filename/PC URI.

PR #1457, opened 2024-06-07
Enable the client to reflect the PC URL used by the server. 

jvwong:
Retrieve the PC_URL from the server endpoint, cache in client.

Refs: #1456

### DIFF
--- a/src/client/features/search/index.js
+++ b/src/client/features/search/index.js
@@ -17,8 +17,6 @@ const { ErrorMessage } = require('../../common/components/error-message');
 const { FeatureView } = require('./feature-view');
 const { Contribute } = require('../../common/components/contribute');
 
-const { PC_URL } = require('../../../config');
-
 class Search extends React.Component {
 
   constructor(props) {
@@ -67,6 +65,9 @@ class Search extends React.Component {
 
   componentDidMount() {
     this.getSearchResult();
+    ServerAPI.getPCURL().then(PC_URL => {
+      this.PC_URL = PC_URL;
+    });
   }
 
   onSearchValueChange(e) {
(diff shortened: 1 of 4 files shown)


jvwong:
This isn't a general solution, e.g. `FACTOID_URL` is a config variable, but works for our purposes. Our React version is pretty old (docs aren't even available) and there are some new features to handle these sorts of things....

IgorRodchenkov (approved):
Looks good, thank you!

PS (somewhat off-topic: 

Before  v4.0.0 release of this app for PC v14 backend (new app-ui will not completely work with the older versions/data anymore due to some env/config changes and using bioregistry.io instead of identifiers.org, etc.) -  

- can we make sure we have correct Dockerfile for prod (I bet we do not need imges for dev use anyway...). Is this line there still correct:  `RUN NODE_ENV=development npm ci` (I guess NODE_ENV is not needed or must be production, as we do not want any dev modules/dependencies in docker images, and it should be `npm install` instead of ci.. but I am not sure)?
- check/polish Readme as well... 
- do we really want that "Syblars" and cli.js be part of this web app or those could be separated into another Git repo perhaps?

IgorRodchenkov:
@jvwong , unlike FACTOID_URL, which is only used by the server-side/backend (client calls webapp's /api/biofactoid proxy  instead of calling biofactoid api directly), PC_URL was used both from backend and client side. There was just one use case - downloading data in varios formats directly from PC ws as saving as files - where PC_URL was used from the client. 
This PR here seems to remove the pre-built/default PC_URL value from the app-ui bundle.js and so the PC_URL environment property set at runtime now applies no only to the backend but also to those client's "download" links!

Issue #1470, opened 2024-08-13
Toolbar hyperlinks are missing 


jvwong:
The Toolbar hyperlinks are missing. They should in fact link to the relevant sections on the homepage.

<img width="1380" alt="Screenshot 2024-08-13 at 11 02 30 AM" src="https://github.com/user-attachments/assets/992f20a0-bb86-4f6e-a739-4af392de5d94">

PR #1471, opened 2024-08-13
Update search toolbar links on retrieval of PC_URL from server.

jvwong:
This change is intended to properly set the PC homepage links on the Search page toolbar. I believe the issue was that setting the server URL (`PC_URL`) on the Search component - rather than state - would not trigger a re-render in React. 

TLDR: The backstory for having the client retrieve the server URL (`getPCURL()`) from the backend in the first place was to be able to build assets in a Docker image.

Refs #1470

### DIFF
--- a/src/client/features/search/index.js
+++ b/src/client/features/search/index.js
@@ -35,7 +35,8 @@ class Search extends React.Component {
       feature: null,
       dataSources: [],
       loading: false,
-      error: null
+      error: null,
+      PC_URL: null
     };
   }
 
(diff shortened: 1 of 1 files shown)


maxkfranz:
Looks good.  Let's merge this in

Note for future:  If the request for the PC URL takes longer than expected (e.g. a slow connection), you could have the link not going anywhere.  This could be mitigated by making the PC URL request earlier than the React mount sequence.  Just noting this in case it comes up in future.  Let's leave the request simple and as-is right now.

Issue #1479, opened 2025-04-22
PC Service at /validate/xref is not available


jvwong:
The app tries to access `/api/pc/xref2uri`, ostensibly to retrieve properly formatted entity `xrefLinks` that are used to create hypertext for entities in the UI (e.g. to UniProt), but the service is returning `Bad Gateway`. The result of this in the app is that no gene information (or links) are shown.

e.g. POST `http://localhost:3000/api/pc/xref2uri`
with JSON body:

```json
{
  "query": {
    "name": "uniprot",
    "localId": "P04637"
  }  
}
```

Actual result is : `Bad Gateway : https://www.pathwaycommons.org/validate/xref`

Expected result is a JSON response with the following:

```
{
  "namespace": "uniprot",
  "uri": "https://bioregistry.io/uniprot:P04637"
}
```
---

Quick and easy solution: manually create these `xreLinks`, which is already done within the `ncbi.js` service:

https://github.com/PathwayCommons/app-ui/blob/70d267bdbf228b1d245546ebe4ed30370519e991/src/server/external-services/ncbi.js#L151-L167

---

Links
- https://github.com/PathwayCommons/app-ui/issues/1128

PR #1480, opened 2025-04-24
Manually create xref links

jvwong:
This quick-fix manually creates xrefs to entities that the Search UI displays as hyperlinks. It bypasses a downed pc.xref2uri web service call, that, in retrospect, appears unnecessary. 

Refs #1479


jvwong:
This seems to work in my hands, so I'm going to pull it in for now ASAP, to get the gene-search feature working.

